PR #1, opened 2020-05-06
Redis consumer

sobychacko:
* Provide a reusable consumer bean.
* Options for redis key, queue or topic.
* Adding tests to veify.


functions/consumer/redis-consumer/src/test/java/org/springframework/cloud/fn/consumer/redis/AbstractRedisConsumerTests.java
```diff
@@ -0,0 +1,53 @@
...
+@DirtiesContext
+public class AbstractRedisConsumerTests {
+
+	@Autowired
+	Consumer<Message<?>> redisConsumer;
+
+	@Autowired
+	RedisConnectionFactory redisConnectionFactory;
+
+	StringRedisTemplate createStringRedisTemplate(RedisConnectionFactory connectionFactory) {
```
artembilan, line 42: There is already such a bean in Spring Boot: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfiguration.java#L62

So, you just need to have it autowired on your test level alongside with that `redisConnectionFactory`.
If we need it after that yet, of course...

functions/consumer/redis-consumer/src/test/java/org/springframework/cloud/fn/consumer/redis/RedisConsumerQueueTests.java
```diff
@@ -0,0 +1,49 @@
...
+
+		//Execute
+		Message<String> message = MessageBuilder.withPayload("hello").build();
+
+		redisConsumer.accept(message);
+
+		Object result = redisTemplate.boundListOps("test-queue").rightPop(5000, TimeUnit.MILLISECONDS);
+		assertThat(result).isNotNull();
+
+		assertThat(result).isEqualTo("hello");
```
artembilan, line 47: Just FYI: assertions can be combined or nested:
```
assertThat(result)
    .isNotNull()
    .isEqualTo("hello");
```
But in your case `.isNotNull()` is really redundant 😄

functions/consumer/redis-consumer/src/test/resources/application.properties
```diff
@@ -0,0 +1,4 @@
+spring.redis.host=${embedded.redis.host}
+spring.redis.port=${embedded.redis.port}
+spring.redis.user=${embedded.redis.user}
+spring.redis.password=${embedded.redis.password}
```
artembilan, line 4: New line in the end of each file.

artembilan:
Merged as https://github.com/spring-cloud/stream-applications/commit/86fc242292d215e777fe35f551c6a167e7fe8a06 after private discussion and fixing the mentioned changed.

PR #48, opened 2020-05-20
Add http-request-function and processor app

### LINKED ISSUE #36: Migrate httpclient processor apps
Convert the current httpclient processor as a function and then generate apps for the binders.

dturanski:
Resolves #36


sobychacko:
@dtur Could you also add a simple `README` on the functions side? See [this PR](https://github.com/spring-cloud/stream-applications/pull/47) where I tried to add some basic info. All it needs is some information about how one may inject it as a standalone function and use it. We can also link to the properties class there.

applications/processor/http-request-processor/src/test/resources/META-INF/spring.binders
```diff
@@ -0,0 +1,2 @@
+integration:\
+org.springframework.cloud.stream.binder.test.TestChannelBinderConfiguration
```
artembilan, line 2: New line in the end of file.

applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
```diff
@@ -0,0 +1,2 @@
+configuration-properties.classes=org.springframework.cloud.stream.app.processor.http.request.HttpRequestProcessorProperties,\
+  org.springframework.cloud.stream.app.processor.http.request.HttpRequestProcessorProperties$Retry
```
artembilan, line 2: Looks like there is no `Retry` class any more.

applications/processor/http-request-processor/src/main/java/org/springframework/cloud/stream/app/processor/http/request/HttpRequestProcessorProperties.java
```diff
@@ -0,0 +1,254 @@
...
+ * @author Waldemar Hummer
+ * @author Mark Fisher
+ * @author Christian Tzolov
+ * @author Artem Bilan
+ * @author David Turanski
+ */
+
+@Validated
+@ConfigurationProperties("http.request.processor")
+public class HttpRequestProcessorProperties implements HttpRequestProperties {
```
artembilan, line 46: Even though you don't create a function as a bean over here, you still can configure `http.request` properties.
So, I don't see anything wrong to have two these configuration classes on the processor and do respective logic to distribute properties combination. This way we indeed won't need a common interface to support and will avoid some confusion in the target projects.

applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
dturanski: Yes. We discussed that. It no longer applies here. The binder already implements Retry.

applications/processor/http-request-processor/src/main/java/org/springframework/cloud/stream/app/processor/http/request/HttpRequestProcessorProperties.java
dturanski: This is a consequence of the boot metadata/doc plugin. The only way to generate all the properties in the README AFAICT.  I tried some other (obvious) combinations.

dturanski: Also:
* The validation rules (e.g., only one of urlExpression, url) require the processor properties to "inherit" or compose with the function properties.  The validation rules are different for function properties (e.g. url is required).
* The same prefix for all properties in the app context is less confusing IMHO

applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
artembilan: So, are you going to fix this then?

applications/processor/http-request-processor/src/main/java/org/springframework/cloud/stream/app/processor/http/request/HttpRequestProcessorProperties.java
artembilan: OK. Thanks. Resolved then!

applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
```diff
@@ -0,0 +1,2 @@
+configuration-properties.classes=org.springframework.cloud.stream.app.processor.http.request.HttpRequestProcessorProperties,\
```
artembilan, line 1:
```suggestion
configuration-properties.classes=org.springframework.cloud.stream.app.processor.http.request.HttpRequestProcessorProperties
```

applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
artembilan:
```suggestion
```

PR #54, opened 2020-06-05
JMS Supplier/Source

sobychacko:


applications/processor/bridge-processor/README.adoc.tmp
```diff
@@ -0,0 +1,11 @@
...
+= Bridge Processor
+
+A processor that bridges the input and ouput by simply passing the incoming payload to the outbound.
+
+=== Payload
+
+Any
+
+//end::ref-doc[]
+null
```
artembilan, line 11: `null` ?
And why this file has a `.tmp` name?

functions/supplier/jms-supplier/src/main/java/org/springframework/cloud/fn/supplier/jms/JmsSupplierProperties.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+/**
+ * Properties for the JMS Supplier.
+ *
+ * @author Gary Russell
+ *
+ */
+@ConfigurationProperties(prefix = "jms")
```
artembilan, line 30: I think this has to go with the `jms.supplier` since there might be something like `jms.consumer` in the same classpath.

PR #55, opened 2020-06-09
Websocket supplier/consumer - source/sink

sobychacko:


sobychacko:
@artembilan Maybe we can refactor it to the reactive version in another iteration?

artembilan:
Sure! We can. Let's address other minor issues and merge!

PR #63, opened 2020-06-16
Tensorflow (functional model) redesign

tzolov:
Based on earlier work
 - [Programming Model for TensorFlow Inference](https://docs.google.com/document/d/1Dxm4B_XACOOXLDOPMpK6T7xaMJF8TqxhYH5tM2ag5cA) 
 - [MindModels](https://tzolov.github.io/mind-model-services)


applications/processor/image-recognition-processor/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+                        <configClass>org.springframework.cloud.stream.app.processor.image.recognition.ImageRecognitionProcessorConfiguration.class</configClass>
+                        <functionDefinition>imageRecognitionFunction</functionDefinition>
+                    </generatedApp>
+
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.springframework.cloud.fn</groupId>
+                            <artifactId>image-recognition-function</artifactId>
+                            <version>${java-functions.version}</version>
+                        </dependency>
```
sobychacko, line 66: Don't we need to include the `image-recognition-processor` dependency here? `image-recognition-function` will come as transitive.

applications/processor/object-detection-processor/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+                        <configClass>org.springframework.cloud.stream.app.processor.object.detection.ObjectDetectionProcessorConfiguration.class</configClass>
+                        <functionDefinition>objectDetection</functionDefinition>
+                    </generatedApp>
+
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.springframework.cloud.fn</groupId>
+                            <artifactId>object-detection-function</artifactId>
+                            <version>${java-functions.version}</version>
+                        </dependency>
```
sobychacko, line 67: Same comment as above. I think we should include the processor dependency here.

applications/processor/image-recognition-processor/pom.xml
tzolov: Not sure what you mean? this is the pom of the `image-recognition-processor`. We can't add itself.  But perhaps i misunderstand the question?

sobychacko: It is ok to add the processor there. Otherwise, the processor configuration will not be included in the generated app.

sobychacko:
PR merged upstream.

PR #66, opened 2020-06-18
Add composite function support for the sources

### LINKED ISSUE #65: Add support for composite functions for all sources
Auto configure reusable functions to make them available for composition with all (most) sources out of the box.

dturanski:
Resolves #65  - Depends on #62  - Actually branched from that branch, so will have to rebase if there are changes.


applications/source/file-source/pom.xml
```diff
@@ -26,6 +26,12 @@
             <artifactId>file-supplier</artifactId>
             <version>${java-functions.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud.stream.app</groupId>
+            <artifactId>stream-applications-composite-function-support</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
```
dturanski, line 34: This isn't strictly needed unless we write some tests, but I included in some of the sources in anticipation and just in case it causes any issues. It did catch a test failure in one source that did not include the function definition.

sobychacko:
Merged upstream.

PR #72, opened 2020-07-01
Add support for s3 compatible storage to s3 consumer and supplier.

tsalm-pivotal:


functions/common/aws-s3-common/pom.xml
```diff
@@ -33,6 +33,13 @@
             <groupId>org.springframework.integration</groupId>
             <artifactId>spring-integration-file</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.8.2</version>
```
artembilan, line 40: We don't use JUnit 4 any more.

functions/common/aws-s3-common/src/test/java/org/springframework/cloud/fn/common/aws/s3/AmazonS3ConfigurationTest.java
```diff
@@ -0,0 +1,60 @@
...
+import com.amazonaws.services.s3.AmazonS3Client;
+import org.junit.Assert;
+import org.junit.Test;
+
+import org.springframework.cloud.aws.core.region.StaticRegionProvider;
+
+/**
+ * @author Timo Salm
+ */
+public class AmazonS3ConfigurationTest {
```
artembilan, line 35: We prefer to name them as `Tests`. Seems for me there were some environments which didn't see test classes without a proper suffix...

functions/consumer/s3-consumer/src/main/java/org/springframework/cloud/fn/consumer/s3/AwsS3ConsumerEndpointConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * @author Timo Salm
+ */
+@Configuration
+public class AwsS3ConsumerEndpointConfiguration {
+
+	@Bean
+	@ConditionalOnProperty("s3.consumer.endpoint-url")
```
artembilan, line 33: Huh? Why this cannot be as a conditional bean in the common `AmazonS3Configuration`?
Why do we need to duplicate it? 
Right we probably need to introduce an `S3ClientProperties`, but then we just don't have any changes in the `supplier`, nor `consumer`.

functions/consumer/s3-consumer/src/test/java/org/springframework/cloud/fn/consumer/s3/AwsS3ConsumerEndpointConfigurationTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.cloud.aws.core.region.RegionProvider;
+import org.springframework.cloud.aws.core.region.StaticRegionProvider;
+import org.springframework.context.annotation.Bean;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Timo Salm
+ */
+class AwsS3ConsumerEndpointConfigurationTest {
```
artembilan, line 34: Why this is not a `Tests` ?

tsalm-pivotal:
I changed the implementation based on your suggestion but I'm not sure where to add the documentation.

tsalm-pivotal:
Both, the sink and the source work with s3 compatible storage (tested with EMC storage).

functions/common/aws-s3-common/src/main/java/org/springframework/cloud/fn/common/aws/s3/CompatibleStorageAmazonS3Configuration.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.cloud.aws.core.region.RegionProvider;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+
+/**
+ * @author Timo Salm
+ */
+@Configuration
+@EnableConfigurationProperties(AmazonS3Properties.class)
```
artembilan, line 35: Mark this class with `@AutoconfigureBefore(AmazonS3Configuration.class)` and you don't need that `@Primary` any more!

tsalm-pivotal: Results in a failing testAmazonS3ConfigurationForS3CompatibleStorage Test ->  `org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'com.amazonaws.services.s3.AmazonS3' available: expected single matching bean but found 2: amazonS3,compatibleStorageAmazonS3`

artembilan: (Please, study how to do multi-line code snippet in GH comments: https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks)

If you see that behavior then something is wrong. 
We definitely should have only a single `AmazonS3` bean.
Probably that `AmazonS3Configuration` comes without a `@ConditionalOnMissingBean`. Time to fix that?

tsalm-pivotal: So it looks like without the @Primary the AmazonS3 Bean created by the spring-cloud-starter-aws dependency will always be the default (and also the AmazonS3 Bean in AmazonS3Configuration will never be used).

artembilan: As far as I know the `spring-cloud-starter-aws` doesn't create `AmazonS3`. Please, elaborate more.
Thanks

artembilan: Turns out the `ContextResourceLoaderAutoConfiguration` creates for us S3 client bean and that one cannot be overridden. Probably time to talk to Spring Cloud AWS team...

functions/common/aws-s3-common/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+spring.autoconfigure.exclude=org.springframework.cloud.aws.autoconfigure.context.ContextResourceLoaderAutoConfiguration
```
artembilan, line 1: Consider to configure your IDE to always insert a new line in the end of all the files.

functions/common/aws-s3-common/src/main/java/org/springframework/cloud/fn/common/aws/s3/AmazonS3Properties.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Timo Salm
+ */
+@ConfigurationProperties("s3.common")
+public class AmazonS3Properties {
+
+	/**
+	 * Optional S3 endpoint url to connect to s3 compatible storage.
```
artembilan, line 28: Let's remove the first `S3` - sounds like tautology. 
More over it probably going to mislead: an `s3://` is really about a resource, not an endpoint connection. This one probably is an `http://` in most cases.

functions/common/aws-s3-common/src/main/java/org/springframework/cloud/fn/common/aws/s3/AmazonS3Properties.java
```diff
@@ -0,0 +1,39 @@
...
+	 */
+	private String endpointUrl;
+
+	public String getEndpointUrl() {
+		return endpointUrl;
```
artembilan, line 33: Always `this.` prefix for properties access.

functions/common/aws-s3-common/src/test/java/org/springframework/cloud/fn/common/aws/s3/AmazonS3ConfigurationTests.java
```diff
@@ -0,0 +1,78 @@
...
+			Assertions.assertEquals(testRegionName, amazonS3.getRegionName());
+			Assertions.assertTrue(amazonS3.getResourceUrl("b", "k")
+					.startsWith("https://s3.eu-central-1.amazonaws.com"));
+		});
+	}
+
+	@Test
+	public void testAmazonS3ConfigurationForS3CompatibleStorage() {
+		runner.withPropertyValues(
+				"s3.common.endpoint-url=https://object.ecstestdrive.com"
```
artembilan, line 58: Looks like this one is a real URL. We probably should consider to use some mock `localhost` to avoid an unexpected connection to the Internet.

artembilan:
@tsalm-pivotal ,

thank you very much for the contribution; looking forward for more!

Issue #75, opened 2020-07-08
Disabling of auto configured AmazonS3 client doesn't work with the s3 stream applications


tsalm-pivotal:
The disabling of the auto-configured AmazonS3 client worked for the functions with the exclusion of the auto-configuration class in  [application.properties](https://github.com/spring-cloud/stream-applications/blob/master/functions/common/aws-s3-common/src/main/resources/application.properties) of the [aws-s3-common](https://github.com/spring-cloud/stream-applications/tree/master/functions/common/aws-s3-common) submodule. It doesn't work for the applications. Probably because Spring Boot doesn't support inheriting properties.

tsalm-pivotal:
The AmazonS3 client is by default auto-configured by spring-cloud-aws-autoconfigure:
https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextResourceLoaderAutoConfiguration.java extends https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-context/src/main/java/org/springframework/cloud/aws/context/config/annotation/ContextResourceLoaderConfiguration.java  where the AmazonS3 client is created.

artembilan:
Yeah... That's what I was afraid of. When we have an `application.properties` in the nested jar of our final Spring Boot application, it is not visible for Spring Boot infrastructure to parse. So, only the way is to use an `EnvironmentPostprocessor` to set a respective property programmatically.

/CC @sobychacko

PR #76, opened 2020-07-08
Fix disabling of auto configured AmazonS3 client

tsalm-pivotal:
Related to issue https://github.com/spring-cloud/stream-applications/issues/75


tsalm-pivotal:
Tested in a live environment!

PR #92, opened 2020-07-16
Support filterFunction and spelFunction composition

dturanski:
Refactor to support composition of filter and spel functions:

Provide `FilterFunctionProperties`  so `filter.function.expression`  and `spel.function.expression` can be specified. 

see `TimeSourceTests`


dturanski:
Squashed and merges

PR #94, opened 2020-07-27
Adjust content type headers for compatibility with SCSt 2.1 apps

dturanski:
Adds auto-configuration of a BPP to add a channel interceptor to function input channels


dturanski:
This fixes the compatibility issue related to the `contentType` header between SCSt 3.0 and SCSt 2.1. The 2.1 version overwrites the `contentType` header on the producer and saves the original value in the `originalContentType` header. The 2.1 consumer restores the `contentType` header. The current binder does not use  the`originalContentType` header, resulting in a type conversion error in some cases.

sobychacko:
LGTM. Squashed and merged upstream.

PR #99, opened 2020-08-07
Add true Consumer and tests

dturanski:


dturanski:
@olegz - Should SCSt should handle the case of declaring a Function in a sink?  The behavior I see without implementing an explicit [Consumer](https://github.com/dturanski/stream-applications/blob/fix-mongodb-sink/functions/consumer/mongodb-consumer/src/main/java/org/springframework/cloud/fn/consumer/mongo/MongoDbConsumerConfiguration.java) in  [this test](https://github.com/dturanski/stream-applications/blob/fix-mongodb-sink/applications/sink/mongodb-sink/src/main/test/org/springframework/cloud/stream/app/sink/mongodb/MongodbSinkTests.java) is that the function is invoked but the message is never converted so nothing is saved when  a `byte[]` is passed to the mongo template.  

cc: @sobychacko @artembilan

sobychacko:
@dturanski @olegz Merging this PR. We should take a look at the issue mentioned [above](https://github.com/spring-cloud/stream-applications/pull/99#issuecomment-670694531) as that pattern may be used in other functions/applications.

PR #101, opened 2020-08-11
Refactor to use SpEL in function properties

dturanski:
This move all SpEL expression evaluation into the function. Originally, it was only supported at the app level.  
Removed dual properties for static and expression values.


functions/function/http-request-function/pom.xml
```diff
@@ -19,6 +19,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.springframework.cloud.fn</groupId>
+            <artifactId>config-common</artifactId>
```
artembilan, line 24: What does this `common` bring for us into this function?

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
```diff
@@ -50,36 +53,65 @@ public WebClient webClient() {
...
+					.bodyValue(resolveBody(message))
+					.headers(httpHeaders -> httpHeaders.addAll(resolveHeaders(message)))
 					.retrieve()
 					.toEntity(properties.getExpectedResponseType())
+					.map(responseEntity -> properties.getReplyExpression().getValue(responseEntity))
 					.timeout(Duration.ofMillis(properties.getTimeout())));
 		}
+
+		private String resolveUrl(Message<?> message) {
+			return properties.getUrlExpression().getValue(message, String.class);
```
artembilan, line 90: Looks like you don't use any `EvaluationContext`, so we don't have access to bean factory from target expressions.

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionProperties.java
```diff
@@ -35,64 +34,64 @@
...
 	 */
-	private HttpMethod httpMethod = DEFAULT_HTTP_METHOD;
+	private Expression urlExpression;
 
 	/**
-	 * The type used to interpret the response.
+	 * A SpEL expression to derive the request method from the incoming message.
 	 */
-	private Class<?> expectedResponseType = DEFAULT_RESPONSE_TYPE;
+	private Expression httpMethodExpression = new SpelExpressionParser().parseExpression("'GET'");
```
artembilan, line 59: We probably can just use here a `new ValueExpresison<>(HttpMethod.GET)`.
No reason to have a hard SpEL parsing against static value.

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionProperties.java
```diff
@@ -35,64 +34,64 @@
...
-	@NotEmpty
-	@Override
-	public String getUrl() {
-		return this.url;
-	}
+	/**
+	 * A SpEL expression used to compute the final result, applied against the whole http
+	 * {@link org.springframework.http.ResponseEntity}.
+	 */
+	private Expression replyExpression = new SpelExpressionParser().parseExpression("body");
```
artembilan, line 75: Can we use a plain `new FunctionExpression<HttpEntity<?>>(HttpEntity::getBody)` for optimization?

functions/function/http-request-function/pom.xml
dturanski: Used to bind Expression properties

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
dturanski: The spelConverter from common-config sets it. https://github.com/spring-cloud/stream-applications/blob/master/functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/SpelExpressionConverterConfiguration.java#L77 Does that work?

artembilan: Yep! It does. Thank you!

PR #105, opened 2020-08-17
Syslog supplier/source

sobychacko:
- Provide TCP/UDP based syslog supplier and then generate apps based on that

### DIFF
--- a/applications/source/syslog-source/pom.xml
+++ b/applications/source/syslog-source/pom.xml
@@ -41,7 +41,7 @@
                         <name>syslog</name>
                         <type>source</type>
                         <version>${project.version}</version>
-                        <configClass>org.springframework.cloud.fn.supplier.syslog.TcpSupplierConfiguration.class</configClass>
+                        <configClass>org.springframework.cloud.fn.supplier.syslog.SyslogSupplierConfiguration.class</configClass>
                     </generatedApp>
                     <dependencies>
                         <dependency>
--- a/applications/source/syslog-source/pom.xml
+++ b/applications/source/syslog-source/pom.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>syslog-source</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+    <name>syslog-source</name>
+    <description>syslog source apps</description>
+    <packaging>jar</packaging>
+
+    <parent>
+        <groupId>org.springframework.cloud.stream.app</groupId>
+        <artifactId>stream-applications-core</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+        <relativePath/>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud.fn</groupId>
+            <artifactId>syslog-supplier</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-app-starter-doc-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.springframework.cloud.stream.app.plugin</groupId>
+                <artifactId>spring-cloud-stream-app-maven-plugin</artifactId>
+                <configuration>
+                    <generatedApp>
+                        <name>syslog</name>
+                        <type>source</type>
+                        <version>${project.version}</version>
+                        <configClass>org.springframework.cloud.fn.supplier.syslog.TcpSupplierConfiguration.class</configClass>
+                    </generatedApp>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.springframework.cloud.fn</groupId>
+                            <artifactId>syslog-supplier</artifactId>
+                        </dependency>
+                        <dependency>
+                            <groupId>org.springframework.cloud.stream.app</groupId>
+                            <artifactId>stream-applications-composite-function-support</artifactId>
+                            <version>${stream-apps-core.version}</version>
+                        </dependency>
+                    </dependencies>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <repositories>
+        <repository>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+            <id>spring-snapshots</id>
+            <name>Spring Snapshots</name>
+            <url>https://repo.spring.io/libs-snapshot-local</url>
+        </repository>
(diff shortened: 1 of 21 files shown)


applications/source/syslog-source/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.springframework.cloud.stream.app.plugin</groupId>
+                <artifactId>spring-cloud-stream-app-maven-plugin</artifactId>
+                <configuration>
+                    <generatedApp>
+                        <name>syslog</name>
+                        <type>source</type>
+                        <version>${project.version}</version>
+                        <configClass>org.springframework.cloud.fn.supplier.syslog.TcpSupplierConfiguration.class</configClass>
```
artembilan, line 44: You said `UDP`, too. So, what happened with that one over here?

applications/source/syslog-source/src/main/resources/META-INF/dataflow-configuration-metadata.properties
```diff
@@ -0,0 +1,2 @@
+configuration-properties.classes=org.springframework.cloud.fn.supplier.syslog.SyslogSupplierProperties
```
artembilan, line 1: What is the difference between this file and the previous?
Can't we do the stuff just with one of them?

functions/supplier/syslog-supplier/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
```
artembilan, line 41: Why do we need `web` for `supplier`?
Especially in the `test` scope...

functions/supplier/syslog-supplier/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-security</artifactId>
```
artembilan, line 46: Same for security.

functions/supplier/syslog-supplier/src/main/java/org/springframework/cloud/fn/supplier/syslog/SyslogSupplierConfiguration.java
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright 2016-2020 the original author or authors.
```
artembilan, line 2: Probably we can treat this class as new. So, no any past in its source code!

functions/supplier/syslog-supplier/src/main/java/org/springframework/cloud/fn/supplier/syslog/SyslogSupplierConfiguration.java
```diff
@@ -0,0 +1,172 @@
...
+import org.springframework.integration.syslog.DefaultMessageConverter;
+import org.springframework.integration.syslog.MessageConverter;
+import org.springframework.integration.syslog.RFC5424MessageConverter;
+import org.springframework.integration.syslog.inbound.RFC6587SyslogDeserializer;
+import org.springframework.integration.syslog.inbound.SyslogReceivingChannelAdapterSupport;
+import org.springframework.integration.syslog.inbound.TcpSyslogReceivingChannelAdapter;
+import org.springframework.integration.syslog.inbound.UdpSyslogReceivingChannelAdapter;
+import org.springframework.messaging.Message;
+
+@Configuration
```
artembilan, line 45: Some `@author` and general JavaDocs?

Also: I see all Spring and Spring Boot configuration classes use a `proxyBeanMethods = false`.
Probably for better performance. See its JavaDocs.
I wonder if we should consider something like that for our configurations as well.

functions/supplier/syslog-supplier/src/main/java/org/springframework/cloud/fn/supplier/syslog/SyslogSupplierProperties.java
```diff
@@ -0,0 +1,130 @@
...
+	public void setRfc(String rfc) {
+		this.rfc = rfc;
+	}
+
+	@AssertTrue(message = "rfc must be 5424 or 3164")
+	public boolean isSupportedRfc() {
+		return "5424".equals(this.rfc) || "3164".equals(this.rfc);
+	}
+
+	@AssertTrue(message = "protocl must be 'udp' or 'tcp' or 'both'")
```
artembilan, line 126: Probably we should consider to make `protocl` as an enum. (hope you see a typo 😉 )
Same for the `rfc`.

functions/supplier/syslog-supplier/src/test/java/org/springframework/cloud/fn/supplier/syslog/AbstractSyslogSupplierTests.java
```diff
@@ -0,0 +1,110 @@
...
+import org.springframework.integration.ip.tcp.connection.AbstractServerConnectionFactory;
+import org.springframework.integration.ip.udp.UnicastReceivingChannelAdapter;
+import org.springframework.integration.syslog.inbound.UdpSyslogReceivingChannelAdapter;
+import org.springframework.messaging.Message;
+import org.springframework.test.annotation.DirtiesContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE, properties = "syslog.supplier.port = 0")
```
artembilan, line 43: I don't think we would need to worry about a `webEnvironment` when we remove `web` and `security` dependencies.

applications/source/syslog-source/src/main/resources/META-INF/dataflow-configuration-metadata.properties
sobychacko: We provide both for SCDF backward compatibility reasons. We are deprecating the file that contains `whitelist`,  but provide here for the time being so that SCDF users can continue to use it on older versions of SCDF.

functions/supplier/syslog-supplier/src/main/java/org/springframework/cloud/fn/supplier/syslog/SyslogSupplierConfiguration.java
sobychacko: good idea. We can probably tackle that across the board separately?

artembilan: Sure! Let's raise a GH issue then!

functions/supplier/syslog-supplier/src/main/java/org/springframework/cloud/fn/supplier/syslog/SyslogSupplierProperties.java
```diff
@@ -125,6 +125,23 @@ public boolean isSupportedRfc() {
 
 	@AssertTrue(message = "protocol must be 'udp' or 'tcp' or 'both'")
 	public boolean isSupportedProtocol() {
-		return "tcp".equals(this.protocol) || "udp".equals(this.protocol) || "both".equals(this.protocol);
+		return "tcp".equals(this.protocol.name()) || "udp".equals(this.protocol.name()) || "both".equals(this.protocol.name());
```
artembilan, line 128: I think this check is pointless: you just can't provide anything else unless a constant from `Protocol`.

PR #112, opened 2020-08-28
Add container base image config

dturanski:
Add container configuration to all app poms. This works with https://github.com/spring-cloud/spring-cloud-dataflow-apps-plugin/pull/2 but should be ok to merge independently.


tzolov:
@dturanski , looks like this change can e applied once inside the core's global section: 
https://github.com/spring-cloud/stream-applications/blob/master/applications/stream-applications-core/pom.xml#L311-L313
?

PR #114, opened 2020-09-15
Migrating Mail Source

sobychacko:
* Mail supplier for receiving email from a URL (IMAP/POP3)
 * Generating mail source apps
 * Migrating tests

### DIFF
--- a/applications/source/mail-source/pom.xml
+++ b/applications/source/mail-source/pom.xml
@@ -1,0 +1,142 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>mail-source</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+    <name>mail-source</name>
+    <description>mail source apps</description>
+    <packaging>jar</packaging>
+
+    <parent>
+        <groupId>org.springframework.cloud.stream.app</groupId>
+        <artifactId>stream-applications-core</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+        <relativePath/>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud.fn</groupId>
+            <artifactId>mail-supplier</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud.stream.app</groupId>
+            <artifactId>stream-applications-composite-function-support</artifactId>
+            <version>${stream-apps-core.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.integration</groupId>
+            <artifactId>spring-integration-test-support</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-dataflow-apps-docs-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.springframework.cloud</groupId>
+                <artifactId>spring-cloud-dataflow-apps-generator-plugin</artifactId>
+                <configuration>
+                    <application>
+                        <name>mail</name>
+                        <type>source</type>
+                        <version>${project.version}</version>
+                        <configClass>org.springframework.cloud.fn.supplier.mail.MailSupplierConfiguration.class</configClass>
+
+                        <maven>
+                            <dependencies>
+                                <dependency>
+                                    <groupId>org.springframework.cloud.fn</groupId>
+                                    <artifactId>mail-supplier</artifactId>
+                                </dependency>
+                                <dependency>
+                                    <groupId>org.springframework.cloud.stream.app</groupId>
+                                    <artifactId>stream-applications-composite-function-support</artifactId>
+                                    <version>${stream-apps-core.version}</version>
+                                </dependency>
+                            </dependencies>
+                        </maven>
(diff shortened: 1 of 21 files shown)


applications/source/mail-source/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+
+                        <maven>
+                            <dependencies>
+                                <dependency>
+                                    <groupId>org.springframework.cloud.fn</groupId>
+                                    <artifactId>mail-supplier</artifactId>
+                                </dependency>
+                                <dependency>
+                                    <groupId>org.springframework.cloud.stream.app</groupId>
+                                    <artifactId>stream-applications-composite-function-support</artifactId>
```
artembilan, line 65: Why do we need this here?
I see it is in a `<scope>test</scope>` for the project.
Anyway: is there a reason that we need this dep at all?
Thanks

applications/source/mail-source/src/test/java/org/springframework/cloud/stream/app/source/file/MailSourceTests.java
```diff
@@ -0,0 +1,90 @@
...
+
+			OutputDestination target = context.getBean(OutputDestination.class);
+			Message<byte[]> sourceMessage = target.receive(10000);
+			final String actual = new String(sourceMessage.getPayload());
+			assertThat(actual.endsWith("\r\n\r\nfoo\r\n\r\n"));
+		}
+	}
+
+	@SpringBootApplication
+	@Import(MailSupplierConfiguration.class)
```
artembilan, line 87: Don't we make them as auto-configuration?
I recall that there was a discussion on the matter...
Where did it go, please?

applications/stream-applications-core/pom.xml
```diff
@@ -425,6 +425,10 @@
                                             <groupId>org.springframework.boot</groupId>
                                             <artifactId>spring-boot-starter-security</artifactId>
                                         </dependency>
+                                        <dependency>
+                                            <groupId>com.jayway.jsonpath</groupId>
+                                            <artifactId>json-path</artifactId>
```
artembilan, line 430: Why do we need this?
Why haven't we needed it before?
What does really make you to bring it now?

functions/supplier/jdbc-supplier/README.adoc
```diff
@@ -1,6 +1,6 @@
-# JDBC Supplier
+# Mail Supplier
```
artembilan, line 1: Probably some mistake... Because this is really in the `jdbc-supplier`

functions/supplier/mail-supplier/README.adoc
```diff
@@ -0,0 +1,34 @@
+# Mail Supplier
+
+This module provides a File supplier that can be reused and composed in other applications.
+The `Supplier` uses the mail IMAP and POP3 support from Spring Integration.
+`MailSupplier` is implemented as a `java.util.function.Supplier`.
```
artembilan, line 5: I don't think that we really have a `MailSupplier` class.
Can we somehow change this text to not mislead end-users?

functions/supplier/mail-supplier/pom.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>mail-supplier</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+    <name>mail-supplier</name>
+    <description>mail supplier</description>
```
artembilan, line 7: Probably need to be Capitalized...

artembilan (changes requested):
Some review

applications/source/mail-source/pom.xml
sobychacko: good point. we don't need that.

applications/source/mail-source/src/test/java/org/springframework/cloud/stream/app/source/file/MailSourceTests.java
sobychacko: I think we decided against it. From what I remember, the main reason was to avoid unnecessary components from getting autoconfigured, but we can revisit that decision. Anycase, I think that discussion is beyond the scope of this PR.

applications/stream-applications-core/pom.xml
sobychacko: You are right, this is not strictly related to this PR, but I found out that the apps were missing this dependency and added it here.

artembilan: OK. But there is no justification of the change in the commit message, so we could be lost bumping to the change in the future.

functions/supplier/jdbc-supplier/README.adoc
```diff
@@ -1,6 +1,6 @@
 # JDBC Supplier
 
-This module provides a JDBC supplier that can be reused and composed in other applications.
+This module provides a Mail supplier that can be reused and composed in other applications.
```
artembilan, line 3: Huh? There is still this `mail` word in JDBC readme...

functions/supplier/mail-supplier/README.adoc
```diff
@@ -0,0 +1,34 @@
+# Mail Supplier
+
+This module provides a File supplier that can be reused and composed in other applications.
+The `Supplier` uses the mail IMAP and POP3 support from Spring Integration.
+`mailSupplier` bean is implemented as a `java.util.function.Supplier`.
```
artembilan, line 5: "The `mailSupplier`..." ?

artembilan (changes requested):

applications/source/mail-source/pom.xml
sobychacko: Thought i addressed that: https://github.com/spring-cloud/stream-applications/pull/114/commits/bda65b0c8a225af01a9f780f3773f7d8af94b561#diff-392bd2c586a7d4c4a7adf45c35896c0aL28

applications/stream-applications-core/pom.xml
sobychacko: Ok, will remove it from the PR.

applications/source/mail-source/pom.xml
sobychacko: Oh, I removed this from the source level dependency for testing, but we are still including it for composition purposes. We are including this dependency on all the sources.

PR #116, opened 2020-09-18
[Do-Not-Merge-Yet]Rsocket consumer

sobychacko:


applications/sink/rsocket-sink/src/test/java/org/springframework/cloud/stream/app/rsocket/sink/RSocketSinkTests.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.handler.annotation.MessageMapping;
+import org.springframework.stereotype.Controller;
+
+public class RSocketSinkTests {
+
+	@Test
+	public void testRsocketSink() throws Exception {
+		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(
```
artembilan, line 41: This test doesn't finish because you just send a data into an `InputDestination`.
There is a function subscribed to that destination like this `Function<Message<?>, Mono<Void>>`.
And looks like nothing subscribes to that `Mono`.

Probably time to talk to Oleg to see what we can do for this Reactive "consumer" to make a subscription automatic.

In MongoDbConsumer David did a hack like:
```
@Bean
public Consumer<Message<?>> mongodbConsumer(Function<Message<?>, Mono<Void>> mongodbConsumerFunction) {
	return message -> mongodbConsumerFunction.apply(message).subscribe();
}
```

Honestly I'm not a fun of this solution and would prefer to have something automatic on the framework side: some `flatMap()` or `handle()` should be done over there to honor back-pressure.

artembilan (changes requested):
Here is my review.

functions/consumer/rsocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/rsocket/RsocketConsumerConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Configuration
+@EnableConfigurationProperties(RsocketConsumerProperties.class)
+public class RsocketConsumerConfiguration {
+
+	@Bean
+	public Function<Message<?>, Mono<Void>> rsocketConsumer(RSocketRequester.Builder builder,
+															RsocketConsumerProperties rsocketConsumerProperties) {
+		final Mono<RSocketRequester> rSocketRequester = builder.connectTcp(rsocketConsumerProperties.getHost(),
+				rsocketConsumerProperties.getPort());
```
artembilan, line 37: If we stick with Spring Boot `2.3` for time being, please, consider to add `cache()` operator to this `Mono`.
Otherwise we are going to reconnect on every single message passed to the function

artembilan (changes requested):

sobychacko:
Thanks @artembilan for the review. As discussed, I am going to send a new PR with the suggested changes.

PR #119, opened 2020-10-05
3.1.0-WIP

dmfrey:
start 3.1.0-WIP branch

all work integrating with Spring Integration 5.4.x should be derived from this branch

### DIFF
--- a/applications/stream-applications-core/pom.xml
+++ b/applications/stream-applications-core/pom.xml
@@ -26,7 +26,7 @@
         <spring-cloud-dataflow-apps-docs-plugin.version>1.0.0-M1</spring-cloud-dataflow-apps-docs-plugin.version>
         <spring-cloud-stream.version>3.1.0-SNAPSHOT</spring-cloud-stream.version>
         <spring-cloud-function-dependencies.version>3.1.0-SNAPSHOT</spring-cloud-function-dependencies.version>
-		    <spring-integration-dependencies.version>5.4.0-SNAPSHOT</spring-integration-dependencies.version>
+        <spring-integration-dependencies.version>5.4.0-SNAPSHOT</spring-integration-dependencies.version>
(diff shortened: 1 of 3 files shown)


applications/stream-applications-core/pom.xml
```diff
@@ -4,31 +4,31 @@
...
         <stream-apps-core.version>${project.version}</stream-apps-core.version>
         <java-functions.version>1.0.0-SNAPSHOT</java-functions.version>
         <spring-cloud-dataflow-apps-generator-plugin.version>1.0.0-M1</spring-cloud-dataflow-apps-generator-plugin.version>
         <spring-cloud-dataflow-apps-docs-plugin.version>1.0.0-M1</spring-cloud-dataflow-apps-docs-plugin.version>
-        <spring-cloud-stream.version>3.0.9.BUILD-SNAPSHOT</spring-cloud-stream.version>
-        <spring-cloud-function-dependencies.version>3.0.9.RELEASE</spring-cloud-function-dependencies.version>
-		<spring-integration-dependencies.version>5.3.3.BUILD-SNAPSHOT</spring-integration-dependencies.version>
+        <spring-cloud-stream.version>3.1.0-SNAPSHOT</spring-cloud-stream.version>
+        <spring-cloud-function-dependencies.version>3.1.0-SNAPSHOT</spring-cloud-function-dependencies.version>
+		    <spring-integration-dependencies.version>5.4.0-SNAPSHOT</spring-integration-dependencies.version>
```
artembilan, line 29: Please , fix this indent. I don't think that SI deserves its own special view over here 😄

artembilan (changes requested):
Please, consider to issue PR against `3.1-WIP` branch in `upstream`, not `master`.
We definitely can't override `master` at the moment because we still develop `3.0` version.
Your effort indeed is going to be applied for the next cycle.

Thanks

artembilan:
Let me know if that doesn't work for you and I'll create `3.1-WIP` branch `upstream`!

PR #122, opened 2020-10-06
Feature/zeromq supplier

dmfrey:
implement a `java.util.function.Supplier` for accepting data from a ZeroMQ TCP server


functions/supplier/zeromq-supplier/mvnw
```diff
@@ -0,0 +1,253 @@
...
+else
+    echo Deactivating \"central\" profile for version=\"$VERSION\"
+    echo $MAVEN_ARGS | grep -q central && MAVEN_ARGS=$(echo $MAVEN_ARGS | sed -e 's/-Pcentral//')
+fi
+
+exec "$JAVACMD" \
+  $MAVEN_OPTS \
+  -classpath "$MAVEN_PROJECTBASEDIR/.mvn/wrapper/maven-wrapper.jar" \
+  "-Dmaven.home=${M2_HOME}" "-Dmaven.multiModuleProjectDirectory=${MAVEN_PROJECTBASEDIR}" \
+  ${WRAPPER_LAUNCHER} ${MAVEN_ARGS} "$@"
```
artembilan, line 253: I think Maven scripts is an overhead: we probably just rely here on the root project level.
So, IMO these scripts must be removed.

functions/supplier/zeromq-supplier/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+            <groupId>org.springframework.integration</groupId>
+            <artifactId>spring-integration-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.integration</groupId>
+            <artifactId>spring-integration-zeromq</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
```
artembilan, line 27: And why do we need Web on the supplier level?

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@ConfigurationProperties("zeromq.supplier")
+@Validated
+public class ZeroMqSupplierProperties {
+
+	private String connectUrl;
+	private String[] topics = {""};
```
artembilan, line 35: Since this is a `@ConfigurationProperties`, some plain JavaDocs must be present on these properties to generate the proper Spring Boot metadata.
Plus we use them to generate README for source app.

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+import org.zeromq.SocketType;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@ConfigurationProperties("zeromq.supplier")
```
artembilan, line 30: Some `@author` and `@since` would be great to have on all the classes you wrote.

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,80 @@
...
+@Validated
+public class ZeroMqSupplierProperties {
+
+	private String connectUrl;
```
artembilan, line 34: I wonder what is wrong with the `bindPort` property as well...
Don't you think that someone may consider to have source app as a server for ZeroMq socket?

functions/supplier/zeromq-supplier/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+            <artifactId>spring-boot-configuration-processor</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-validation</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
```
artembilan, line 40: Doesn't this come transitively from the `spring-boot-starter-validation` ?

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @param topics The ZeroMQ Topics to subscribe to
+	 *
+	 * @see org.springframework.integration.zeromq.inbound.ZeroMqMessageProducer#setTopics(String...)
+	 */
+	public void setTopics(String... topics) {
+		this.topics = topics;
+	}
+
+	public SocketType getReceiveSocketType() {
+		return SocketType.SUB;
```
artembilan, line 69: Why isn't this property exposed?
Why be so explicit with its name?
Why just `socketType` is not enough?

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
artembilan: Consider to expose also:
```
	/**
	 * Specify a {@link Duration} to delay consumption when no data received.
	 * @param consumeDelay the {@link Duration} to delay consumption when empty;
	 *                     defaults to {@link #DEFAULT_CONSUME_DELAY}.
	 */
	public void setConsumeDelay(Duration consumeDelay) {
```

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+
+	private FluxMessageChannel output = new FluxMessageChannel();
+
+	@Bean
+	public ZContext zContext() {
+		return new ZContext();
+	}
+
+	@Bean
+	public ZeroMqMessageProducer adapter(ZContext zContext) {
```
artembilan, line 53: We probably need to have a `Consumer<ZMQ.Socket> socketConfigurer` as optional dependency.
End-user might need some socket options to provide, like security or read timeout options.
I'm not sure that we can expose all the socket options as a configuration properties (see `SocketBase.setSocketOpt(int option, Object optval)`), but would be great to think about some general.

artembilan (changes requested):
Here is some review so far.

Thank you for the contribution, @dmfrey !

functions/supplier/zeromq-supplier/mvnw
dmfrey: @artembilan I agree. I was attempting to remain consistent with the other projects. I'll remove them. The root `.mvnw` or an installed maven distribution allows you to target any submodule with `-pl :submodule_name`.

functions/supplier/zeromq-supplier/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+        <groupId>org.springframework.cloud.fn</groupId>
+        <artifactId>spring-functions-parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../../spring-functions-parent</relativePath>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.integration</groupId>
+            <artifactId>spring-integration-core</artifactId>
```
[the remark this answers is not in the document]
dmfrey: removing it

functions/supplier/zeromq-supplier/pom.xml
dmfrey: removing it

functions/supplier/zeromq-supplier/pom.xml
dmfrey: removing it

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
dmfrey: will add :)

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
dmfrey: @artembilan  I was expecting this supplier to only connect to a server and receive messages from it. Should it also be able to configure a server? I guess I was expecting the consumer equivalent (forthcoming) to do that.

dmfrey: Will add the `consumeDelay` as a configurable option

artembilan: No, `consumer` is just a publisher and to not let it to lose messages it is better to not have it responsible for a void state on the server side. At least that's my feeling about ZeroMq socket since we don't have a persistence in between. 
My point is here that with ZeroMq supplier as a server you kind introduce a middleware between ZeroMq publishers and persistent binder. So, if no server backed by our supplier, no middleware to produce into the binder!

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
dmfrey: Will expose it but default it to SUB

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfiguration.java
dmfrey: How about I expose the `socketConfigurer` as a `@Bean` so that a user could come in and supply their own version if they would need to override it?

dmfrey: Do something like:

```
	@Bean
	@ConditionalOnMissingBean
	public Consumer<ZMQ.Socket> socketConfigurer() {
		return (socket) -> { };
	}
```
Then inject it into the adapter bean creation

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,153 @@
...
+	 * @param consumeDelay the {@link Duration} to delay consumption when empty;
+	 *                     defaults to {@link #DEFAULT_CONSUME_DELAY}.
+	 */
+	@NotNull(message = "'consumeDelay' must not be null")
+	public void setConsumeDelay(Duration consumeDelay) {
+		this.consumeDelay = consumeDelay;
+	}
+
+	@NotNull(message = "topics(s) are required")
+	@Size(min = 1, message = "At least one topic is required")
```
artembilan, line 138: That's wrong. If socket type is not SUB, we don't need topics to subscribe.

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	/**
+	 * @param bindPort The TCP Port to bind to on all interfaces
+	 *
+	 * @see org.springframework.integration.zeromq.inbound.ZeroMqMessageProducer#setBindPort(int)
+	 */
+	@Range(min = 0, message = "'bindPort' must not be zero or negative")
+	public void setBindPort(int bindPort) {
+		this.bindPort.set(bindPort);
```
artembilan, line 120: I don't think we need an `AtomicInteger` on the configuration properties level...
It is just an internal logic of the channel adapter.

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,153 @@
...
+	 */
+	private String[] topics = {""};
+
+	/**
+	 * @param socketType the {@link SocketType} to establish,
+	 *                      defaults to {@link #DEFAULT_SOCKET_TYPE} if empty
+	 */
+	@NotNull(message = "'socketType' is required")
+	public void setSocketType(SocketType socketType) {
+		Assert.state(VALID_SOCKET_TYPES.contains(socketType),
```
artembilan, line 84: You don't need this logic here. It is just a matter of channel adapter functionality and it is validated over there.

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @author Daniel Frey
+ * @since 3.1.0
+ */
+@ConfigurationProperties("zeromq.supplier")
+@Validated
+public class ZeroMqSupplierProperties {
+
+	private static final SocketType DEFAULT_SOCKET_TYPE = SocketType.SUB;
+	private static final Duration DEFAULT_CONSUME_DELAY = Duration.ofSeconds(1);
```
artembilan, line 45: I don't think we need all of these constants for default. Just return their values from getters directly!

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,153 @@
...
+
+	private final AtomicInteger bindPort = new AtomicInteger();
+
+	/**
+	 * The {@link SocketType} the connection should make.
+	 *
+	 * Allowable values:
+	 * - PAIR
+	 * - PULL
+	 * - SUB
```
artembilan, line 57: Please, consider to make all of these JavaDocs for props as one line and no JavaDocs tags, please.
See Spring Boot docs: https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.html#configuration-metadata

> A short description of the property that can be displayed to users. If no description is available, it may be omitted. It is recommended that descriptions be short paragraphs, with the first line providing a concise summary. The last line in the description should end with a period (.).

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,153 @@
...
+	public Duration getConsumeDelay() {
+		return consumeDelay;
+	}
+
+	/**
+	 * @param bindPort The TCP Port to bind to on all interfaces
+	 *
+	 * @see org.springframework.integration.zeromq.inbound.ZeroMqMessageProducer#setBindPort(int)
+	 */
+	@Range(min = 0, message = "'bindPort' must not be zero or negative")
```
artembilan, line 118: It can be `0` and we should treat it as a "random port binding"

artembilan (changes requested):

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+	private FluxMessageChannel output = new FluxMessageChannel();
+
+	@Bean
+	public ZContext zContext() {
+		return new ZContext();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public Consumer<ZMQ.Socket> socketConfigurer() {
```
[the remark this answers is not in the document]
dmfrey: removed, and checked for null before setting

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+	@Bean
+	public ZeroMqMessageProducer adapter(ZeroMqSupplierProperties properties, ZContext zContext, Consumer<ZMQ.Socket> socketConfigurer) {
+
+		ZeroMqMessageProducer zeroMqMessageProducer = new ZeroMqMessageProducer(zContext, properties.getSocketType());
+
+		if (properties.getConnectUrl() != null) {
+			zeroMqMessageProducer.setConnectUrl(properties.getConnectUrl());
+		}
+		else {
+			zeroMqMessageProducer.setBindPort(properties.getBoundPort());
```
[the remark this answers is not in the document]
dmfrey: @artembilan  added if condition to check for greater than 0. Is 0 acceptable here or should it at least be 1024, since you can't typically bind to a port lower than 1024 on most systems (without requiring root access)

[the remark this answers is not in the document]
artembilan: Well, let's defer that to the OS! If it is `0`, we don't set and rely on the random port. If it is `> 0`, we set and done: everything else is out of our control.

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
dmfrey: removed validation check. added check when setting topics to ensure the socketType is SUB

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfiguration.java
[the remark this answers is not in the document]
artembilan: Please, don't comment if you agree with me: I'll see changes in the upcoming commits.
Otherwise it looks like a spam.
That's me sending you a single message with the whole review, but comments come as a separate emails.
So, destructive 😄 

Thanks for understanding!

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
artembilan: No, that's not. The topic can be empty with the meaning "subscriber to all". See `ZMQ.SUBSCRIPTION_ALL`

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -41,48 +41,36 @@
...
 	/**
 	 * The {@link SocketType} the connection should make.
-	 *
-	 * Allowable values:
-	 * - PAIR
-	 * - PULL
-	 * - SUB
 	 */
-	private SocketType socketType = DEFAULT_SOCKET_TYPE;
+	private SocketType socketType;
```
artembilan, line 47: You probably want to set `SocketType.SUB` as a default value here...

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -41,48 +41,36 @@
...
-	 * Mutually exclusive with the {@link #bindPort}.
+	 * Connection URL for to the ZeroMQ TCP Server.
 	 */
 	private String connectUrl;
 
 	/**
-	 * A {@link Duration} to delay consumption when no data received.
+	 * Bind Port for connecting to the ZeroMQ TCP Server.
+	 */
+	private Integer bindPort;
```
artembilan, line 57: Can't it be just primitive `int` with the default meaning as `0`?
Of course, we should just ignore its value at all when `connectUrl` is provided.
Or you can also validate that port is not `> 0` when `connectUrl` is provided...

artembilan (changes requested):
Sorry, I don't see tests for this supplier.
I thought they've been here before...

functions/supplier/zeromq-supplier/src/test/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfigurationTests.java
```diff
@@ -0,0 +1,89 @@
...
+
+		String socketAddress = "tcp://*";
+		ZMQ.Socket socket = CONTEXT.createSocket(SocketType.PUB);
+		int bindPort = socket.bindToRandomPort(socketAddress);
+
+		ZeroMqSupplierProperties properties = new ZeroMqSupplierProperties();
+		properties.setConnectUrl("tcp://localhost:" + bindPort);
+		properties.setTopics("test-topic");
+
+		ZeroMqSupplierConfiguration configuration = new ZeroMqSupplierConfiguration();
```
artembilan, line 62: I don't think this is what we want to have in test.
Probably really better to have as much Spring context as possible.

Consider to spawn server socket in the `@BeforeAll` and store its bound port into the `System.setProperty()` and then refer to that property from the `@SpringBootTest`.

So, we will have the whole coverage: for functionality and configuration properties binding.

artembilan (changes requested):

functions/supplier/zeromq-supplier/src/test/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfigurationTests.java
```diff
@@ -39,37 +42,41 @@
  * @author Daniel Frey
  * since 3.1.0
  */
+@SpringBootTest(properties = {"zeromq.supplier.topics=test-topic"})
```
artembilan, line 45: Add `@DirtiesContext`, please, so our application context is destroyed properly.

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -64,7 +63,6 @@
 	/**
 	 * @param socketType the {@link SocketType} to establish.
 	 */
-	@NotNull(message = "'socketType' is required")
```
artembilan: You probably wanted to move this onto getter. 
The point is that property like this `zeromq.supplier.socket-type=` is not acceptable.
Same for `consumeDelay`

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,129 @@
...
+	 */
+	private SocketType socketType = SocketType.SUB;
+
+	/**
+	 * Connection URL for to the ZeroMQ TCP Server.
+	 */
+	private String connectUrl;
+
+	/**
+	 * Bind Port for connecting to the ZeroMQ TCP Server.
```
artembilan, line 49: We probably need to mention that "0 means random".
Plus it looks like your description is misleading: we don't connect to the server - we really create a server with bind port.
In fact it doesn't look like in ZeroMQ socket API there is a concept like client/server: there is connecting socket and bound one. Who is the server or client is a target application decision.

(Sorry for being for nasty, but I foresee some questions from end-users which would be better to avoid if we are good with our docs)

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierProperties.java
```diff
@@ -0,0 +1,129 @@
...
+	 */
+	private int bindPort;
+
+	/**
+	 * The delay to consume from the ZeroMQ TCP Server when no data received.
```
artembilan, line 54: Let's don't talk about TCP at all! In fact we can connect via any other supported protocol in ZeroMq!
(Plus "server" remark again...)

functions/supplier/zeromq-supplier/src/main/java/org/springframework/cloud/fn/supplier/zeromq/ZeroMqSupplierConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+
+	private FluxMessageChannel output = new FluxMessageChannel();
+
+	@Bean
+	public ZContext zContext() {
+		return new ZContext();
+	}
+
+	@Bean
+	public ZeroMqMessageProducer adapter(ZeroMqSupplierProperties properties, ZContext zContext, @Autowired(required = false) Consumer<ZMQ.Socket> socketConfigurer) {
```
artembilan, line 54: How about to `socketConfigurer` param to a new line to make the code more readable?
Kinda, let's don't overflow 120 code line length!

artembilan (changes requested):

PR #125, opened 2020-10-09
refactor versions from pom files

dmfrey:
refactor hardcoded versions in `functions` submodule branches

This is so they can be updated in a minimal amount of places and leverage the maven hierarchy to propagate them where appropriate. While working with SNAPSHOT releases of Spring Boot, Spring Integration, and Spring Cloud, it required touching every `pom.xml` file in order to reflect the use of the SNAPSHOT. After this change, it will only be required to change the property in the top `functions` pom.xml and the versions will reflect to all submodules.

Maven introduced the ability use a key property, `revision` to propagate this information to the `<parent>` declarations in each pom.

This was enabled in Maven 3.5+
https://maven.apache.org/maven-ci-friendly.html

### DIFF
--- a/functions/common/aws-s3-common/pom.xml
+++ b/functions/common/aws-s3-common/pom.xml
@@ -1,17 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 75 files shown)


dturanski:
@dmfrey  - The main concern is that we intentionally separated `spring-functions-parent` from `java-functions-parent` because it is possible to implement a function without spring dependencies.  This doesn't apply to the functions we have written since we rely heavily on spring integration and boot autoconfig. But the idea is that we wanted to allow the community to use and/or contribute non-spring functions.   Another thing we traditionally haven't done is to include any dependencies in the parent pom. Only dependency management.  I don't recall the rationale for this but @sobychacko may have an opinion. 
Other than that, LGTM.

sobychacko:
@dmfrey Per our discussions merged the PR and did further restructuring. Thank you very much for doing this work!

PR #133, opened 2020-11-02
Feature/zeromq sink fix

dmfrey:
@artembilan this should be a cleaner PR now. Did as instructed, created new branch from the 3.1-WIP branch and added back in only the added files from the original PR.


applications/sink/zeromq-sink/README.adoc
```diff
@@ -0,0 +1,41 @@
...
+=== Payload
+
+N/A
+
+== Options
+
+The **$$zeromq$$** $$sink$$ has the following options:
+
+//tag::configuration-properties[]
+//end::configuration-properties[]
```
artembilan, line 21: Why is this section empty?
How have you generated them before for the `source` app?

applications/sink/zeromq-sink/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.junit.vintage</groupId>
+                    <artifactId>junit-vintage-engine</artifactId>
```
artembilan, line 34: You know, it looks like we don't need to exclude this dep any more.
It is only like this: `testRuntimeOnly("org.junit.vintage:junit-vintage-engine")`. Therefore is not visible for us at all.

applications/sink/zeromq-sink/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+            <name>Spring Milestones</name>
+            <url>https://repo.spring.io/libs-milestone-local</url>
+        </repository>
+        <repository>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+            <id>spring-releases</id>
+            <name>Spring Releases</name>
+            <url>https://repo.spring.io/release</url>
```
artembilan, line 107: It is recommended now do not use this repo.
Just rely on Maven Central!

applications/sink/zeromq-sink/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+        </repository>
+        <repository>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+            <id>spring-libs-release</id>
+            <name>Spring Libs Release</name>
+            <url>https://repo.spring.io/libs-release</url>
```
artembilan, line 115: Or this...
I would remove both though.

applications/sink/zeromq-sink/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+            <id>spring-milestones</id>
+            <name>Spring Milestones</name>
+            <url>https://repo.spring.io/libs-milestone-local</url>
+        </pluginRepository>
+    </pluginRepositories>
+
+</project>
```
artembilan, line 150: Please, be sure that we have a new line in the end of each file.

applications/sink/zeromq-sink/src/test/java/org/springframework/cloud/stream/app/sink/zeromq/ZeroMqSinkTests.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Tests for ZeroMqSink.
+ *
+ * @author Daniel Frey
+ * @since 3.1.0
+ */
+public class ZeroMqSinkTests {
+
+    private static final ZContext CONTEXT = new ZContext();
```
artembilan, line 46: We have to destroy `ZContext` after all tests:
```
	@AfterAll
	static void tearDown() {
		CONTEXT.close();
	}
```

artembilan (changes requested):
That's all. nothing critical.
We are very close to the merge when we agree those minor points.

applications/sink/zeromq-sink/pom.xml
dmfrey: I also removed the `libs-release` in the plugin dependencies as well.

applications/sink/zeromq-sink/README.adoc
dmfrey: Not sure. Originally, I copied in one of the other files and modified it for this project. I also checked the plugin configuration and it seems to be correct unless I'm missing something in its configuration.

artembilan: Ok. Did you try `install` ? 
@sobychacko , what Maven command should we use to make props available in readers like this?
Thanks

applications/sink/zeromq-sink/src/main/resources/dataflow-configuration-metadata-whitelist.properties
```diff
@@ -0,0 +1 @@
+configuration-properties.classes=org.springframework.cloud.fn.consumer.zeromq.ZeroMqConsumerProperties
```
artembilan, line 1: Both of these files have to be in the `/META-INF` directory.

See all other apps.
The README should be generated properly.

artembilan (changes requested):

PR #137, opened 2020-12-17
Bulk indexing

klinamen:
Support for batch indexing of documents through the ES bulk API was added.
The implementation follows the lines of jdbc-consumer, where the consumer is written as an IntegrationFlow that makes use of the AggregatorMessageHandler from Spring Integration.
Differently from jdbc-consumer, payloads are aggregated into a collection of wrappers of Message. I found no way to customize the Aggregator's default behavior of splitting collections of messages, thus using a collection of wrapped  messages allowed for preserving the individual message headers, which was required to keep the possibility of passing the document id via headers.
I hope you guys can suggest a cleaner solution.

### DIFF
--- a/functions/consumer/elasticsearch-consumer/src/main/java/org/springframework/cloud/fn/consumer/elasticsearch/ElasticsearchConsumerConfiguration.java
+++ b/functions/consumer/elasticsearch-consumer/src/main/java/org/springframework/cloud/fn/consumer/elasticsearch/ElasticsearchConsumerConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2020 the original author or authors.
+ * Copyright 2020-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,6 +59,10 @@ import org.springframework.messaging.MessageHandler;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
+/**
+ * @author Soby Chacko
+ * @author Andrea Montemaggio
+ */
 @Configuration
 @EnableConfigurationProperties(ElasticsearchConsumerProperties.class)
 public class ElasticsearchConsumerConfiguration {
--- a/functions/consumer/elasticsearch-consumer/src/main/java/org/springframework/cloud/fn/consumer/elasticsearch/ElasticsearchConsumerConfiguration.java
+++ b/functions/consumer/elasticsearch-consumer/src/main/java/org/springframework/cloud/fn/consumer/elasticsearch/ElasticsearchConsumerConfiguration.java
@@ -17,12 +17,20 @@
 package org.springframework.cloud.fn.consumer.elasticsearch;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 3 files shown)


sobychacko:
@klinamen I am looking to merge this PR. Did you run against an Elasticsearch cluster and verify the changes you made? Also, could you please rebase the PR against the latest master and update the PR?

klinamen:
Thank for your time in reviewing this PR. I have tried it against an actual ES instance and it works as expected.
However, I am not entirely sure how bulk indexing behaves in case of some of the index operation fails for some of the documents. Do you think is it something we should care about at the sink implementation level, or we can rely on the mechanisms already provided at the binding level?

Looking at the code after a while, I would say we should check the result for each of the documents sent for indexing and handle failed items properly (how? any idea?). I do not remember I have seen anything similar in the jdbc-consumer implementation, and this concern may also apply to it, due to the use of the aggregator to buffer messages.

Before we go ahead, I would like to know your thoughts on that, and try to add a test case for this situation.

artembilan:
I'm not sure why JDBC comes to the question. With that `batchSize` option on JDBC Consumer we are going to commit all records to DB or nothing. IF ES works differently here, then we probably can consider different approach to handle errors...

klinamen:
For implementing bulk indexing on ES I borrowed the integration pattern from JDBC consumer operating in batch mode, and my concern on proper error handling may apply, regardless of the specific data store involved.
Essentially, here we are in the situation in which the commit of a batch to the data store may fail (totally or partially) after successfully consuming from the underlying message broker. Thus, even if we throw an exception, only the message that triggered the release of the batch (which is, by definition, outside the failing batch) will be subject to the error handling policy configured for the binder, and the entire batch will be lost regardless of the configured policy (e.g. retry, DLQ).
Is this a real concern or I am missing something?

artembilan:
Well, that's the price we would pay for such a logic.

In fact it goes even further: if you configure an `idleTimeout`, the group release happens on a different thread and your consumer on the binder level fully lost even with the last message.

So, or we make a decision for ourselves how to live with it or we need to go deeper and revise such a solution in favor of plain consume-n-store transaction for a single message.

On the other hand I just might not see the whole picture and we need a review from someone else in the team...

klinamen:
@artembilan I totally agree. We could leave these implementations as they are, and just clearly document the limitation of the current batch model, so that the user can make an informed decision.

sobychacko:
@klinamen That sounds good to me based on what @artembilan suggested. Please feel free to update the docs and then rebase the PR against the current master. Then I will take it for a final spin before merging. Thanks!

sobychacko:
@klinamen Any chance you can rebase this PR to the master branch?

klinamen:
@sobychacko I have rebased to master. However, I would like to throw on partial failures in bulk indexing operation and add a test case for that. I do not think I'll be able to do so until next weekend, though.

sobychacko:
@klinamen Thank you for the quick response. Please take your time to address those concerns.

klinamen:
@sobychacko, finally I've got some time to clean-up this PR and it is ready for review.
I've added another small feature: now it is possible to override the destination index via a new message header.
Thank you for your patience.

sobychacko:
@klinamen Thanks for the updates and this PR. This is a significant improvement to the Elasticsearch consumer. Do you mind updating the copyright years on the java classes you touched and adding an author tag in the comments with your name on it?

klinamen:
@sobychacko Just pushed the updates. I also added an author tag with your name as the original author of this module.

sobychacko:
@klinamen Thank you for this PR! The changes are now merged upstream.

PR #140, opened 2021-01-12
User-specified sorting of remote file list in sftp-supplier

klinamen:
Entries of the remote file list are returned in an unspecified order, making it difficult to implement use-cases where sequential processing of the remote files is required.
This PR introduces basic support for user-specified sorting of remote file lists via a new `SftpSupplierProperties.sortBy` property. Currently, only a predefined set of properties of `LsEntry` can be used for sorting (filename, last access time, last modified time), which are defined by the `SortSpec.Field` enumeration.


dturanski:
@klinamen  - Thank you for the contribution.  I changed `field` to `attribute`.  Merged by 65f3b82d

klinamen:
@dturanski You are welcome. Totally agree on renaming to `attribute`.

PR #141, opened 2021-01-13
Add WF tracing support

tzolov:
- Add dependency for WF boot starter and s-c-sleuth-starter.
 - Update the common app properties to include wf.application=${stream.name} and wf.service=${app.name}-${app.type}
 - Update the common app properties to wf.tracing.enabled to false by default.

 Related to https://github.com/spring-cloud/spring-cloud-dataflow/issues/3967


sobychacko:
Looks good, merged upstream.

PR #144, opened 2021-01-29
Use Spring Test Data Geode instead of containerized geode server

upthewaterspout:
Use spring test data geode in the geode integration tests
test to fork a separate geode process, instead of running
a geode server in a docker container.

STDG is faster than launching geode through gfsh inside of a container.
- It doesn't run in docker
- It doesn't start management components necessary for gfsh
- It doesn't start extra gfsh shell processes

Removing the GeodeContainer test utility


pivotal-issuemaster:
@upthewaterspout Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/stream-applications&pullRequestId=144)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/stream-applications&pullRequestId=144) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@upthewaterspout Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/stream-applications&pullRequestId=144)!

artembilan (changes requested):
@upthewaterspout ,

thank you for your effort, but I don't think it matters here too much.
Spring Data Geode project uses its own solution because it is high enough in the dependency tree. This project is in the end of the chain where we have access to all the features of Spring Cloud. Saying that, our goal here is to have as less code as possible and concentrate on the business features for this project and don't support an extra code for testing. Since Testcontainers gives us a good support for many services we should test against, we just sacrifice performance in favor of maintainability. 

Plus we have some experience on CI server, when forking JVM in the code fails from time to time. Such a distribution of responsibility to Docker gives more stable build history.

Just my five fingers...

upthewaterspout:
I understand not wanting any flakiness in your CI! 

I noticed some other data store tests were using an embedded data store or a forked process rather than containers. Cassandra and mongodb were the ones I looked at. So I thought it might be reasonable to speed up the Geode tests by doing the same. But if you prefer testing with a containerized data store no worries.

artembilan:
Well, embedded Cassandra and MongoDB don't fork JMV, they run their own process. We may reconsider to do that over Testcontainers as well, but that's different story since they have good embedded support unlike this one for Geode.

dturanski:
Just catching up. I requested the PR after our brief slack thread. I've been on PTO and will review it this week. I do share some of Artem's concerns and appreciate the contribution.

dturanski:
@upthewaterspout - Thanks for the contribution! We will keep an eye on any flakiness and revert to TC if necessary.  Merged via 9faa2a6a

PR #153, opened 2021-02-26
Add the possibilities to configure the headersMappedLast for the RabbitMQ consumer

### LINKED ISSUE #152: Add the possibilities to configure the headersMappedLast for the RabbitMQ consumer
Hello,

The RabbitMQ consumer is using by default a `org.springframework.amqp.support.converter.SimpleMessageConverter` that is setting the content type when `headersMappedLast` is set to false (default).

It means when the payload is of type `byte[]` the content type will be set to `application/octet-stream` whereas in most case it should remain unchanged: RabbitMQ is in most of the case a passthrough medium. In a scenario involving 2 streams connected by a queue, with 1 RabbitMQ sink and 1 RabbitMQ source, it can create issue as the app next to the source is receiving `application/octet-stream` and not the right `application/json`. 

Please let me know if it makes sense, I can then submit a PR.

Nicolas

nithril:
Fix #152

### DIFF
--- a/applications/sink/rabbit-sink/src/test/java/org/springframework/cloud/stream/app/sink/rabbit/HeadersMappedLastTest.java
+++ b/applications/sink/rabbit-sink/src/test/java/org/springframework/cloud/stream/app/sink/rabbit/HeadersMappedLastTest.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2016-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.app.sink.rabbit;
+
+import org.junit.jupiter.api.Test;
+
+import org.springframework.amqp.core.Message;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.test.context.TestPropertySource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class HeadersMappedLastTest {
+
+	@TestPropertySource(properties = { "rabbit.routingKey=scsapp-testq",
+			"rabbit.headersMappedLast=true" })
+	public class HeadersMappedLastTrueTests extends RabbitSinkIntegrationTests {
+
+		@Test
+		public void test() {
+			this.channels.send(MessageBuilder.withPayload("{}".getBytes())
+					.setHeader("content-type", "application/json")
+					.build());
+			this.rabbitTemplate.setReceiveTimeout(10000);
+			Message received = this.rabbitTemplate.receive("scsapp-testq");
+			assertThat(new String(received.getBody())).isEqualTo("{}");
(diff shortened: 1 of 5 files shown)


functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerProperties.java
```diff
@@ -73,6 +73,18 @@
...
 	private boolean ownConnection;
 
+	/**
+	 * When mapping headers for the outbound message, determine whether the headers are
+	 * mapped before the message is converted, or afterwards. This only affects headers
+	 * that might be added by the message converter. When false, the converter's headers
+	 * win; when true, any headers added by the converter will be overridden (if the
+	 * source message has a header that maps to those headers). For example, when using a
+	 * SimpleMessageConverter with a 'byte[]' payload that contains json; if headersMappedLast
+	 * is false, the converter will set the content type to 'application/octet-stream'.
```
artembilan, line 83: This is too long for this type of javadocs.
Such a property is used as a Spring Boot metadata information and is populated for code completion assistant in IDEs (for example). So, it is really recomened to have such a description as short as possible.
More over we have respective documentation is Spring Integration on the matter: https://docs.spring.io/spring-integration/docs/current/reference/html/amqp.html#content-type-conversion-outbound. Therefore no need to be excessive in this project and its results.

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerProperties.java
```diff
@@ -142,4 +154,11 @@ public void setOwnConnection(boolean ownConnection) {
 		this.ownConnection = ownConnection;
 	}
 
+	public boolean isHeadersMappedLast() {
+		return headersMappedLast;
```
artembilan, line 158: `this.` must be presented to satisfy our code style.

applications/sink/rabbit-sink/src/test/java/org/springframework/cloud/stream/app/sink/rabbit/HeadersMappedLastTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import org.springframework.amqp.core.Message;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.test.context.TestPropertySource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class HeadersMappedLastTest {
```
artembilan, line 27: We prefer to name test classes with a `*Tests` suffix.
Please, also add your name to the `@author` of this list and have a copyright of this class update to the current `2021` year.

applications/sink/rabbit-sink/src/test/java/org/springframework/cloud/stream/app/sink/rabbit/HeadersMappedLastTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@TestPropertySource(properties = { "rabbit.routingKey=scsapp-testq",
+			"rabbit.headersMappedLast=true" })
```
artembilan, line 30: We probably don't need this property here since it is `true` by default now. So, we just would test it like "and what it is a default behavior?"

artembilan (changes requested):
Just some minor remarks.

Thank you for your effort so far!

artembilan (approved):
:+1:

nithril:
Thank you for your review and merge

artembilan:
Thank you for your contribution!
Looking forward for more!

Issue #156, opened 2021-03-10
sftp Application Not Respecting Polling Intervals


fiidim:
Prior to 3.0.0 versions of sftp application, `trigger.fixed-delay` was exposed.  This had the effect of polling the sftp channel at a time interval.  In 3.0.x versions, this parameter is gone, and a new parameter called `supplier.delay-when-empty` exists.  When using this in a similar fashion, the sftp polling doesn't delay and continues to poll frantically.  The parameter set I'm using is:

```
supplier.filename-pattern='*.dat'
supplier.factory.username='user'
supplier.factory.password='password'
supplier.factory.allow-unknown-keys=true
supplier.factory.host=x.x.x.x
supplier.remote-dir=/data/outgoing
supplier.local-dir=/data/incoming
supplier.delete-remote-files=true
supplier.delay-when-empty=15s
supplier.stream=true
file.consumer.mode=lines
```

I expected the 2 parameters to behave the same way.  That doesn't seem to be the case.  Am I doing something wrong?

artembilan:
There is not trigger any more for this type of application, because it is not *pollable*, but *reactive*. So, no delay in between SFTP dir reads - only that `delay-when-empty` if no new files to read.

I wonder why your config comes without an `sftp.` prefix?..

artembilan:
Works for me:
```
13:56:16.269 [sshd-SshServer[355ce81c]-nio2-thread-9] INFO  o.a.s.s.s.ServerUserAuthService - Session user@/127.0.0.1:52842 authenticated
13:56:16.304 [boundedElastic-3] INFO  com.jcraft.jsch - Disconnecting from localhost port 52837
13:56:16.304 [Connect thread localhost session] INFO  com.jcraft.jsch - Caught an exception, leaving main loop due to Socket closed
13:56:31.315 [boundedElastic-3] INFO  com.jcraft.jsch - Connecting to localhost port 52837
13:56:31.315 [boundedElastic-3] INFO  com.jcraft.jsch - Connection established
```

You probably need to use the version `3.0.1` of the SFTP Source application.

artembilan:
I'm not sure that it related somehow...
May you try with the `3.0.2-SNAPSHOT` then?

I can't reproduce and confirm that with these options:
```
sftp.supplier.delay-when-empty=15s
sftp.supplier.stream=true
file.consumer.mode=lines
```

I step into a `Mono.delay()` for `15 seconds` in debug mode.

fiidim:
[...]

PR #164, opened 2021-04-08
GH-156: Fix NPE in the `RemoteFileDeletingAdvice`

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/156

The `MessageSource.receive()` may produce `null`:
The `MessageSourceMutator` impl must honor such an input

* Add `MonoProcessor<Boolean> subscriptionBarrier` to delay
subscription to the source `Flux` until subscription happens
to the supplier's flux.
This way we don't have unexpected interaction with the source when
there are regular endpoints in the flow in between

### DIFF
--- a/applications/source/sftp-source/src/test/java/org/springframework/cloud/stream/app/source/sftp/SftpSourceTests.java
+++ b/applications/source/sftp-source/src/test/java/org/springframework/cloud/stream/app/source/sftp/SftpSourceTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2020 the original author or authors.
+ * Copyright 2020-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


garyrussell:
Why is deletion being done here at all?

Seems wrong to me - it should be deferred until after the file has been processed.

artembilan:
Good point, Gary!

If even that would be just a regular `MethodInterceptor` on the whole `doPoll()` (which includes a message sending to the downstream channel), we still cannot be sure that message is produced to the binder in the same thread.

Anyway it is a different story to investigate...
Probably something like `AcknowledgmentCallback` header would be a way to have in the function implementation for end-user consideration. And then binder would benefit from it to be called when producing to the output destination is done (or not).

dturanski:
@artembilan  - So you think we should merge this and address Gary's point later?

artembilan:
I think so. We have a bug currently, so let's just fix it and stop holding community from using a feature!

The concern Gary is raising has a place, but it is not clear how to address having a chelange in front of us to keep a function signature. So, better to raise an issue and think more how the fix could look.

dturanski:
Merged via 04beb9ec

PR #166, opened 2021-05-01
sftp-supplier: renaming remote files after processing

klinamen:
Certain integration scenarios require moving remote files to a different path after processing, rather than deleting them.
This PR adds this capability to sftp-supplier through the new 'sftp-supplier.rename-remote-files-to' property, which a SpEl expression resolving to the new name of the file at hand.

### DIFF
--- a/functions/common/file-common/src/main/java/org/springframework/cloud/fn/common/file/remote/RemoteFileRenamingAdvice.java
+++ b/functions/common/file-common/src/main/java/org/springframework/cloud/fn/common/file/remote/RemoteFileRenamingAdvice.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.fn.common.file.remote;
+
+import java.util.Objects;
+
+import org.springframework.expression.Expression;
+import org.springframework.integration.aop.MessageSourceMutator;
+import org.springframework.integration.core.MessageSource;
+import org.springframework.integration.file.FileHeaders;
+import org.springframework.integration.file.remote.RemoteFileTemplate;
+import org.springframework.lang.Nullable;
+import org.springframework.messaging.Message;
+
+
+/**
+ * A {@link MessageSourceMutator} that renames a remote file on success.
+ *
+ * @author Andrea Montemaggio
+ *
+ */
+public class RemoteFileRenamingAdvice implements MessageSourceMutator {
+
+	private final RemoteFileTemplate<?> template;
+	private final String remoteFileSeparator;
+	private final Expression newName;
+
+	/**
+	 * Construct an instance with the provided template and separator.
+	 * @param template the template.
+	 * @param remoteFileSeparator the separator.
+	 * @param newNameExp the SpEl expression for the new name.
+	 */
+	public RemoteFileRenamingAdvice(RemoteFileTemplate<?> template,
+									String remoteFileSeparator,
+									Expression newNameExp) {
+		this.template = template;
+		this.remoteFileSeparator = remoteFileSeparator;
+		this.newName = newNameExp;
+	}
+
+	@Nullable
+	@Override
+	public Message<?> afterReceive(@Nullable Message<?> result, MessageSource<?> source) {
+		if (result != null) {
+			String remoteDir = (String) result.getHeaders().get(FileHeaders.REMOTE_DIRECTORY);
+			String remoteFile = (String) result.getHeaders().get(FileHeaders.REMOTE_FILE);
+			String newName = Objects.requireNonNull(this.newName.getValue(result)).toString();
+			this.template.rename(remoteDir + this.remoteFileSeparator + remoteFile, newName);
+		}
+		return result;
+	}
+}
--- a/functions/common/file-common/src/main/java/org/springframework/cloud/fn/common/file/remote/RemoteFileRenamingAdvice.java
+++ b/functions/common/file-common/src/main/java/org/springframework/cloud/fn/common/file/remote/RemoteFileRenamingAdvice.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.fn.common.file.remote;
 
-import java.util.Objects;
-
 import org.springframework.expression.Expression;
 import org.springframework.integration.aop.MessageSourceMutator;
 import org.springframework.integration.core.MessageSource;
@@ -36,7 +34,9 @@ import org.springframework.messaging.Message;
 public class RemoteFileRenamingAdvice implements MessageSourceMutator {
 
 	private final RemoteFileTemplate<?> template;
+
 	private final String remoteFileSeparator;
+
 	private final Expression newName;
 
 	/**
@@ -59,8 +59,10 @@ public class RemoteFileRenamingAdvice implements MessageSourceMutator {
 		if (result != null) {
 			String remoteDir = (String) result.getHeaders().get(FileHeaders.REMOTE_DIRECTORY);
 			String remoteFile = (String) result.getHeaders().get(FileHeaders.REMOTE_FILE);
-			String newName = Objects.requireNonNull(this.newName.getValue(result)).toString();
-			this.template.rename(remoteDir + this.remoteFileSeparator + remoteFile, newName);
+			String newNameValue = this.newName.getValue(result, String.class);
+			if (newNameValue != null && !newNameValue.isEmpty()) {
+				this.template.rename(remoteDir + this.remoteFileSeparator + remoteFile, newNameValue);
+			}
 		}
 		return result;
 	}
--- a/functions/supplier/sftp-supplier/pom.xml
+++ b/functions/supplier/sftp-supplier/pom.xml
@@ -32,6 +32,11 @@
 
         <dependency>
             <groupId>org.springframework.cloud.fn</groupId>
+            <artifactId>config-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud.fn</groupId>
             <artifactId>file-common</artifactId>
             <version>${revision}</version>
         </dependency>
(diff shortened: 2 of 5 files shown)


functions/common/file-common/src/main/java/org/springframework/cloud/fn/common/file/remote/RemoteFileRenamingAdvice.java
```diff
@@ -0,0 +1,67 @@
...
+ * A {@link MessageSourceMutator} that renames a remote file on success.
+ *
+ * @author Andrea Montemaggio
+ *
+ */
+public class RemoteFileRenamingAdvice implements MessageSourceMutator {
+
+	private final RemoteFileTemplate<?> template;
+	private final String remoteFileSeparator;
+	private final Expression newName;
```
artembilan, line 40: Better to wrap any class member with blank lines: much easier to read the code.

functions/common/file-common/src/main/java/org/springframework/cloud/fn/common/file/remote/RemoteFileRenamingAdvice.java
```diff
@@ -0,0 +1,67 @@
...
+		this.newName = newNameExp;
+	}
+
+	@Nullable
+	@Override
+	public Message<?> afterReceive(@Nullable Message<?> result, MessageSource<?> source) {
+		if (result != null) {
+			String remoteDir = (String) result.getHeaders().get(FileHeaders.REMOTE_DIRECTORY);
+			String remoteFile = (String) result.getHeaders().get(FileHeaders.REMOTE_FILE);
+			String newName = Objects.requireNonNull(this.newName.getValue(result)).toString();
```
artembilan, line 62: We prefer to use Spring's `Assert.hasText()`, but still: is it so important here?
Why my expression can't return `null`, for example with a meaning "do not rename".
WDYT?

functions/supplier/sftp-supplier/pom.xml
```diff
@@ -30,6 +30,11 @@
             <artifactId>spring-integration-sftp</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework.cloud.fn</groupId>
+            <artifactId>config-common</artifactId>
```
artembilan, line 35: I guess we need this for `SpelExpressionConverterConfiguration`.
Can  we consider apply it then on the `SftpSupplierConfiguration`, but not only in the test?
I mean without such an import the final function app is not going to work just because you have already provided those `Expression` configuration properties. So, better to keep it working.

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierProperties.java
```diff
@@ -75,6 +74,11 @@
 	 */
 	private boolean deleteRemoteFiles = false;
 
+	/**
+	 * A SpEL expression resolving to the new name remote files must be renamed to after successful transfer.
+	 */
+	private Expression renameRemoteFilesTo = null;
```
artembilan, line 80: I think this is "per file", so better to call the property as `renameRemoteFileTo` and also mention a single file in the JavaDoc for this property.

functions/supplier/sftp-supplier/src/test/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierApplicationTests.java
```diff
@@ -235,6 +240,46 @@ void deleteRemoteFiles() {
 
 	}
 
+	@Test
+	void renameRemoteFilesStream() {
+		defaultApplicationContextRunner
+				.withPropertyValues(
+						"sftp.supplier.stream=true",
+						"sftp.supplier.delete-remote-files=false",
+						"sftp.supplier.rename-remote-files-to='/sftpTarget/' + headers.get('file_remoteFile')")
```
artembilan, line 249: It can be simpler: `headers.file_remoteFile`. There is a special `MapPropertyAccessor` to treat the dot operator and the name after it as a `Map` key to call its `get()`.

functions/supplier/sftp-supplier/src/test/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierApplicationTests.java
```diff
@@ -399,8 +444,8 @@ private void deleteNewSource(Path newSource) throws IOException {
 	}
 
 	@SpringBootApplication
+	@Import(SpelExpressionConverterConfiguration.class)
```
artembilan, line 447: As I said in the comment for the POM change, this must go onto the `SftpSupplierConfiguration`.
The point is even without your change we still have a problem on the `SftpSupplierProperties.Factory.knownHostsExpression` property.
So, your contribution via this PR is really great!

functions/supplier/sftp-supplier/src/test/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierApplicationTests.java
```diff
@@ -399,8 +444,8 @@ private void deleteNewSource(Path newSource) throws IOException {
...
 	static class SftpSupplierTestApplication {
 
 	}
-
```
artembilan, line 405: As I said: every class member has to be surrounded with blank lines, even before the last `}` in the class.

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierConfiguration.java
```diff
@@ -221,6 +226,12 @@ public RemoteFileDeletingAdvice remoteFileDeletingAdvice(SftpRemoteFileTemplate
 			return new RemoteFileDeletingAdvice(sftpTemplate, sftpSupplierProperties.getRemoteFileSeparator());
 		}
 
+		@Bean
+		@ConditionalOnExpression("environment['sftp.supplier.rename-remote-files-to'] != null && environment['sftp.supplier.delete-remote-files'] == 'false'")
```
artembilan, line 230: We don't need so complex condition: you cover mutual exclusion in the configuration properties validation.
So, this one probably could be just `@ConditionalOnProperty(prefix = "sftp.supplier", value = "rename-remote-file-to")`

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierConfiguration.java
```diff
@@ -240,16 +251,24 @@ public RemoteFileDeletingAdvice remoteFileDeletingAdvice(SftpRemoteFileTemplate
...
 									.subscriberContext(
 											context -> (context.put(IntegrationReactiveUtils.DELAY_WHEN_EMPTY_KEY,
 													sftpSupplierProperties.getDelayWhenEmpty())))),
-					fileConsumerProperties)
-					.toReactivePublisher();
+					fileConsumerProperties);
+
+			if (renameRemoteFileHandler != null) {
+				flowBuilder.publishSubscribeChannel(pubsub ->
+						pubsub.subscribe(subFlow -> subFlow.handle(renameRemoteFileHandler).nullChannel())
```
artembilan, line 267: Why do we need here?
Doesn't look like we have a removal logic for `NonStreaming` variant...

artembilan (changes requested):
I think in general looks good.
Just need answers to a couple of questions and some code clean up for consistency.

Thank you for your effort so far!

functions/common/file-common/src/main/java/org/springframework/cloud/fn/common/file/remote/RemoteFileRenamingAdvice.java
klinamen: Agreed. There is no reason to consider null an invalid value. I'll fix that.

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierProperties.java
klinamen: I have considered that, but then I preferred being consistent with `deleteRemoteFiles` naming convention and wrote the JavaDoc that way on purpose. Any thoughts on that?

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierConfiguration.java
klinamen: The removal logic is provided by SftpInboundAdapter for the non-streaming case, and as AFAIK there is no support for the renaming behavior we need in Spring Integration.

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierProperties.java
artembilan: Right. I also would just call that `deleteRemoteFile`, but the bird is out of cage already.
Plus that one is definitely for all unconditionally, when "rename" is per file.
But yeah: that's just my opinion - no strict requirements.

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierConfiguration.java
artembilan: Ah! Right, this one: `.deleteRemoteFiles(sftpSupplierProperties.isDeleteRemoteFiles())`. 
Sorry, good catch!

But yeah: rename is provided via gateway you do. So, all good. 👍

functions/supplier/sftp-supplier/src/test/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierApplicationTests.java
klinamen: `SpelExpressionConverterConfiguration` should be imported automatically by auto-configuration, so the fact it is in the POM should be enough. This also happens in other modules (cfr. sftp-supplier, elasticsearch-consumer, wavefront-consumer, and analytics-consumer) that explicitly import `SpelExpressionConverterConfiguration` only in test classes as the auto-configuration does not apply to them.

artembilan: OK. I missed the fact that it is present as an auto-configuration entry.
Why then it doesn't work in the test since I see the `@SpringBootApplication`?

klinamen: It does actually work. Removed.

klinamen:
@artembilan I think I have addressed your remaining concerns.

artembilan (approved):

artembilan:
@klinamen , 

Thank you for contribution; looking forward for more!

klinamen:
@artembilan You're welcome! Now that we have this merged, I would move on porting this feature is to the ftp-supplier.

Issue #160, opened 2021-03-30
MQTT Source does not listen on defined topics


sobychacko:
For more details, see this: https://github.com/spring-cloud-stream-app-starters/mqtt/issues/18

sabbyanandan:
Probably a similar fix on the tcp-source via https://github.com/spring-cloud/stream-applications/issues/159 applies here also.

rjbgaspar:
Hi  @sobychacko and @sabbyanandan I just checked with version 3.0.1 and the problem still persists,

The client is able to get a connection

**tail -f /mosquitto/log/mosquitto.log**
```
1621841349: New client connected from 172.18.0.1:63296 as scdfClient (p2, c1, k60, u'admin').
```

**Instance mqtt-in.mosquito-inbound logs**
```
    org.springframework.boot.actuate.autoconfigure.web.mappings.MappingsEndpointAutoConfiguration

    org.springframework.boot.autoconfigure.availability.ApplicationAvailabilityAutoConfiguration

    org.springframework.boot.autoconfigure.info.ProjectInfoAutoConfiguration



2021-05-24 07:29:09.624  INFO [mqtt-source,,,] 8054 --- [           main] s.c.s.a.m.s.k.MqttSourceKafkaApplication : Started MqttSourceKafkaApplication in 22.328 seconds (JVM running for 23.836)
```

**Instances for app mqtt-in.log logs**
```
==> standard input <==
```

sobychacko:
@rjbgaspar Sorry about that. We will have a release this week and hopefully, a fix for this will be available in it.

leanadvocate:
> Hi @sobychacko and @sabbyanandan I just checked with version 3.0.1 and the problem still persists,
> 
> The client is able to get a connection
> 
> **tail -f /mosquitto/log/mosquitto.log**
> 
> ```
> 1621841349: New client connected from 172.18.0.1:63296 as scdfClient (p2, c1, k60, u'admin').
> ```
> 
> **Instance mqtt-in.mosquito-inbound logs**
> 
> ```
>     org.springframework.boot.actuate.autoconfigure.web.mappings.MappingsEndpointAutoConfiguration
> 
>     org.springframework.boot.autoconfigure.availability.ApplicationAvailabilityAutoConfiguration
> 
>     org.springframework.boot.autoconfigure.info.ProjectInfoAutoConfiguration
> 
> 
> 
[...]

PR #170, opened 2021-05-26
Refactor the usage of output channel in suppliers

sobychacko:
Redesigning suppliers that use reactive output channel name as
the literal "output" explicitly, as this  cuases some conflicts
with the output channel name usage in Spring Cloud Stream.
The fix is done either by simply renaming the channel name
or by avoid using it directly, but delegate its use through the SI Java DSL.

Resolves https://github.com/spring-cloud/stream-applications/issues/160


artembilan:
See relevant issue on SCSt side: https://github.com/spring-cloud/spring-cloud-stream/issues/2132

PR #177, opened 2021-07-06
Update security config for boot 2.5

dturanski:
Enables anonymous access to `health`, `info`, and `bindings`


sobychacko:
Looks good! Merged upstream.

Issue #183, opened 2021-08-12
is there a way to attach certificate files and token information in MQTT Source


leanadvocate:
Hi Team, 

To subscribe messages in MQTT server we do use certificates and tokens like below:
--cafile ./certificates/mqtt_ca.crt 
--cert ./certificates/mqtt_client.crt 
--key ./certificates/mqtt_client.key 
and Token= XXXXXXXX

Is there any way to pass the above information in SCDF kafka streams 

Thanks
Mathan

artembilan:
The MQTT Source is based on Spring Integration MQTT module: https://docs.spring.io/spring-integration/docs/current/reference/html/mqtt.html#mqtt.
The one uses an Eclipse Paho MQTT Client library. 
It is not fully mapped to what you show, but looks like their `org.eclipse.paho.client.mqttv3.MqttConnectOptions` has an option like:
```
	/**
	 * Sets the SSL properties for the connection.
	 * <p>
	 * Note that these properties are only valid if an implementation of the Java
	 * Secure Socket Extensions (JSSE) is available. These properties are
	 * <em>not</em> used if a SocketFactory has been set using
	 * {@link #setSocketFactory(SocketFactory)}. The following properties can be
	 * used:
	 * </p>
	 * <dl>
	 * <dt>com.ibm.ssl.protocol</dt>
	 * <dd>One of: SSL, SSLv3, TLS, TLSv1, SSL_TLS.</dd>
	 * <dt>com.ibm.ssl.contextProvider
	 * <dd>Underlying JSSE provider. For example "IBMJSSE2" or "SunJSSE"</dd>
	 *
	 * <dt>com.ibm.ssl.keyStore</dt>
	 * <dd>The name of the file that contains the KeyStore object that you want the
	 * KeyManager to use. For example /mydir/etc/key.p12</dd>
	 *
	 * <dt>com.ibm.ssl.keyStorePassword</dt>
	 * <dd>The password for the KeyStore object that you want the KeyManager to use.
	 * The password can either be in plain-text, or may be obfuscated using the
	 * static method:
	 * <code>com.ibm.micro.security.Password.obfuscate(char[] password)</code>. This
	 * obfuscates the password using a simple and insecure XOR and Base64 encoding
	 * mechanism. Note that this is only a simple scrambler to obfuscate clear-text
	 * passwords.</dd>
	 *
	 * <dt>com.ibm.ssl.keyStoreType</dt>
[...]

PR #199, opened 2021-11-15
GH-183: Expose MQTT SSL properties

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/183

* Add `MqttProperties.sslProperties` common MQTT configuration property as a `Map` structure
* Delegate a content fo the `MqttProperties.sslProperties` into `mqttConnectOptions.setSSLProperties()`
in the common `MqttConfiguration`
* Verify that SSL properties are applied in both `MqttSupplierTests` & `MqttConsumerTests`
* Document a new `MqttProperties.sslProperties` configuration property for both consumer and supplier
* Use `eclipse-mosquitto` Docker image for MQTT tests instead of RabbitMQ plugin: the Mosquitto starts much faster

### DIFF
--- a/functions/common/mqtt-common/pom.xml
+++ b/functions/common/mqtt-common/pom.xml
@@ -11,7 +11,7 @@
 
     <artifactId>mqtt-common</artifactId>
     <name>mqtt-common</name>
-    <description>file consumer</description>
+    <description>MQTT Common</description>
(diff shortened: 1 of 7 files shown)


sobychacko:
LGTM! Merged upstream.

PR #203, opened 2021-11-20
Fix more CDC tests

artembilan:


artembilan:
I cannot push directly to `upstream/main`, so creating this PR and merging it immediately.
Thank you for understanding! 😄

Issue #207, opened 2021-12-22
Source Http / Supplier failing due to post without Content-Type header


Zekkenn:
Hi everyone! I was testing the HTTP source with no headers and no body, this raises a null pointer, is this expected?

Making a little research, it seems that the bug is here https://github.com/spring-cloud/stream-applications/blob/b2a29269815045dd3426679db1968265df3bdd1b/functions/supplier/http-supplier/src/main/java/org/springframework/cloud/fn/supplier/http/HttpSupplierConfiguration.java#L68

It tries to parse to string a header with a null value to compare it with another, just I wanted to know if this is expected, and if is this the case, do you mind giving me some highlights about how to extend the code to accept those requests?

This was the local curl that I did: `curl --location --request POST 'http://localhost:29710/akamai'`

This is the stacktrace:

```2021-12-22 10:18:15.178 ERROR [http-source,840cbeae7893dc6f,840cbeae7893dc6f] 30216 --- [     parallel-1] a.w.r.e.AbstractErrorWebExceptionHandler : [763c00fe]  500 Server Error for HTTP POST "/akamai"

org.springframework.integration.transformer.MessageTransformationException: Failed to transform Message in bean 'httpSupplierFlow.header-enricher#1' for component 'httpSupplierFlow.org.springframework.integration.config.ConsumerEndpointFactoryBean#0'; defined in: 'org.springframework.cloud.fn.supplier.http.HttpSupplierConfiguration'; from source: 'bean method httpSupplierFlow'; nested exception is org.springframework.messaging.MessageHandlingException: Expression evaluation failed: org.springframework.cloud.fn.supplier.http.HttpSupplierConfiguration$$Lambda$628/1263634860@1e4d3ce5; nested exception is java.lang.NullPointerException, failedMessage=GenericMessage [payload={}, headers={http_requestMethod=POST, http_requestUrl=http://localhost:29710/akamai, b3=840cbeae7893dc6f-c3dbc71e317e7548-0, nativeHeaders={}, id=ec0c4ebb-8fc5-6821-cf8b-43cdcf0f4898, timestamp=1640186295107}]
	at org.springframework.integration.transformer.MessageTransformingHandler.handleRequestMessage(MessageTransformingHandler.java:125) ~[spring-integration-core-5.5.4.jar!/:5.5.4]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
[...]

PR #208, opened 2021-12-22
GH-207: Fix NPE in the HttpSupplierConfiguration

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/207

When no HTTP `Content-Type` header, the NPE is thrown from the `enrichHeaders()`
endpoint

* Fix NPE extracting header value via `get(MessageHeaders.CONTENT_TYPE, MediaType.class)`
* Make `httpHeaderMapper` bean as a `@ConditionalOnMissingBean` to let end-users
to provide their own custom `HeaderMapper<HttpHeaders>`
* Move `httpSupplierProperties.getMappedRequestHeaders()` setting to the `httpHeaderMapper` bean
* Use `httpHeaderMapper` bean injection into the `WebFlux.inboundChannelAdapter()` instead of
just `httpSupplierProperties.getMappedRequestHeaders()`
* Mention custom `HeaderMapper<HttpHeaders>` in the README

### DIFF
--- a/functions/supplier/http-supplier/README.adoc
+++ b/functions/supplier/http-supplier/README.adoc
@@ -24,6 +24,8 @@ All configuration properties are prefixed with `http`.
 
 For more information on the various options available, please see link:src/main/java/org/springframework/cloud/fn/supplier/http/HttpSupplierProperties.java[HttpSupplierProperties].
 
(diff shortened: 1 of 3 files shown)


sobychacko:
Looks good. Merged upstream.

PR #209, opened 2022-01-19
Change `doOnSubscribe` to `doOnRequest`

artembilan:
The `aggregatorFunction` has a flaw subscribing and emitting
data from a source `Flux` too early: from a `doOnSubscribe()`
callback which is really called *before* the real subscription is
registered in the target `Publisher`

Change the logic to the `doOnRequest()` which happens already
after subscription is registered in the `Publisher`

**Cherry-pick to `2020.0.x`**


sobychacko:
Looks good. Merged upstream.

sobychacko:
Backported to `2020.0.x`.

PR #211, opened 2022-01-24
Restore embedded MongoDb tests

artembilan:


functions/consumer/mongodb-consumer/src/test/java/org/springframework/cloud/fn/consumer/mongo/MongoDbConsumerApplicationTests.java
```diff
@@ -43,8 +42,8 @@
  */
 @SpringBootTest(properties = {
 		"spring.data.mongodb.port=0",
-		"mongodb.consumer.collection=testing" })
-@Disabled
+		"mongodb.consumer.collection=testing",
+		"spring.mongodb.embedded.version=latest" })
```
onobc, line 46: My one concern on using [latest](https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo/blob/c478eb5aa8cd3372ae4be0f8bb5eafa4d821ad7a/src/main/java/de/flapdoodle/embed/mongo/distribution/Version.java#L236) is that it could introduce unexpected errors as they bump this to more advanced server versions. 

SB 2.6.3 recommends Mongo client version 4.4.1. I would prefer to lock onto a similar version of the embedded server such as [4.4.1](https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo/blob/c478eb5aa8cd3372ae4be0f8bb5eafa4d821ad7a/src/main/java/de/flapdoodle/embed/mongo/distribution/Version.java#L228) but those fail to resolve from Flapdoodle at test runtime. I know that [4.0.12](https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo/blob/c478eb5aa8cd3372ae4be0f8bb5eafa4d821ad7a/src/main/java/de/flapdoodle/embed/mongo/distribution/Version.java#L223) resolves. 

Another interesting tidbit is that when I made this change in SB we stayed locked on [3.5.5](https://github.com/spring-projects/spring-boot/pull/27108) for its tests. We should probably bump that up in SB later as well.

artembilan:
I suggest to merge this and rebase #213 .
That one is already too long to consider MongoDB test there as well.

onobc (approved):

PR #212, opened 2022-01-24
Enable authorized access to POST to actuator endpoints

### LINKED ISSUE #210: Allow authorized user to post to `/actuator/bindings`
Related to https://github.com/spring-cloud/spring-cloud-dataflow/issues/4758  and (private) https://github.com/pivotal/scdf-pro-ui/issues/69  We need the ability to grant a user with admin privileges to post to exposed actuator endpoints, notably `bindings`.

dturanski:
Resolves #210

### DIFF
--- a/applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebFluxSecurityAutoConfiguration.java
+++ b/applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebFluxSecurityAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2020 the original author or authors.
+ * Copyright 2019-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,20 +18,32 @@ package org.springframework.cloud.stream.app.security.common;
 
 import reactor.core.publisher.Flux;
 
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.autoconfigure.endpoint.web.WebEndpointProperties;
 import org.springframework.boot.actuate.autoconfigure.security.reactive.ReactiveManagementWebSecurityAutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
+import org.springframework.boot.autoconfigure.security.SecurityProperties;
 import org.springframework.boot.autoconfigure.security.reactive.ReactiveSecurityAutoConfiguration;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.http.HttpMethod;
 import org.springframework.security.config.annotation.web.reactive.EnableWebFluxSecurity;
 import org.springframework.security.config.web.server.ServerHttpSecurity;
+import org.springframework.security.core.userdetails.MapReactiveUserDetailsService;
+import org.springframework.security.core.userdetails.User;
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.security.core.userdetails.UserDetailsService;
+import org.springframework.security.crypto.factory.PasswordEncoderFactories;
 import org.springframework.security.web.server.SecurityWebFilterChain;
 import org.springframework.security.web.server.WebFilterChainProxy;
+import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcher;
+import org.springframework.util.AntPathMatcher;
+import org.springframework.util.StringUtils;
 import org.springframework.web.reactive.config.WebFluxConfigurer;
 
 /**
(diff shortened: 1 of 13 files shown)


applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebFluxSecurityAutoConfiguration.java
```diff
@@ -49,19 +61,61 @@
...
 		if (!securityProperties.isEnabled()) {
 			http.authorizeExchange()
 					.anyExchange()
 					.permitAll();
 		}
 		else {
-			http.authorizeExchange().pathMatchers("/actuator", "/actuator/health", "/actuator/info", "/actuator/bindings")
+			http.authorizeExchange()
+					.pathMatchers(HttpMethod.POST, postToActuatorPath).hasRole("ADMIN")
+					.pathMatchers("/actuator", "/actuator/health", "/actuator/info", "/actuator/bindings")
```
artembilan, line 118: Don't we need to use the same `EndpointRequest.to("health", "info", "bindings")` over here, too?
I guess that `webEndpointProperties.getBasePath()` could be used otherwise, but I believe to keep it as an `/actuator` prefix is wrong.

artembilan (changes requested):

applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebSecurityAutoConfiguration.java
```diff
@@ -47,29 +59,71 @@
...
+
 	@Bean
 	WebSecurityConfigurerAdapter appStarterWebSecurityConfigurerAdapter(
 			AppStarterWebSecurityAutoConfigurationProperties securityProperties) {
 
 		return new WebSecurityConfigurerAdapter() {
-
 			@Override
 			protected void configure(HttpSecurity http) throws Exception {
+				AntPathRequestMatcher postToActuator = new AntPathRequestMatcher(
```
onobc, line 102: One caveat of building up the path like this is that it does not handle the case where the servlet container is configured w/ a context path other than `/` (eg. '/myapp`). The matcher that EndpointRequest [ultimately leverages](https://github.com/spring-projects/spring-boot/blob/4494d3a805fa686b7dc6c56f22d5a2dd1ad978c8/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/security/servlet/SecurityRequestMatchersManagementContextConfiguration.java#L53) does respect the path. 

If we care about that we can do something like I did in my prototype [here](https://github.com/onobc/playground/blob/main/actuator-security/src/main/java/com/example/actuator/security/WebSecurityConfig.java#L48). However, me and @dturanski discussed this and it is probably an edge case that we would not need to worry about.

applications/stream-applications-core/stream-applications-security-common/src/test/java/org/springframework/cloud/stream/app/security/common/AbstractSecurityCommonTests.java
```diff
@@ -33,8 +39,16 @@
 	@Autowired
 	protected TestRestTemplate restTemplate;
 
+
+
 	@SpringBootApplication
+	@Import(TestChannelBinderConfiguration.class)
```
onobc, line 45: Nice add on the test!

applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebFluxSecurityAutoConfiguration.java
onobc: This is a good point. The imperative version does not have the "actuator" paths. It would be a simple change to bring this inline w/ the mvc version. The api does require a string path and we can inject in PathMappedEnpdoints bean do something like this: 
```java
pathMappedEndpoints.getPath(EndpointId.of("health"))
```
to get each path.

applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebSecurityAutoConfiguration.java
```diff
@@ -47,29 +59,71 @@
...
+	AuthenticationProvider adminAuthenticationProvider(
+			AppStarterWebSecurityAutoConfigurationProperties securityProperties) {
+		return new AuthenticationProvider() {
+			@Override
+			public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+				if (securityProperties.isEnabled() &&
+						StringUtils.hasText(securityProperties.getAdminUser()) &&
+						securityProperties.getAdminUser().equals(authentication.getName()) &&
+						StringUtils.hasText(securityProperties.getAdminPassword()) &&
+						securityProperties.getAdminPassword().equals(authentication.getCredentials().toString())) {
```
onobc, line 79: Will `credentials` always be non-null?

applications/stream-applications-core/stream-applications-security-common/src/test/java/org/springframework/cloud/stream/app/security/common/ReactiveSecurityEnabledAuthorizedAccessTests.java
```diff
@@ -40,14 +42,19 @@
 @TestPropertySource(properties = {
 		"spring.main.web-application-type=reactive",
 		"management.endpoints.web.exposure.include=*",
+		"management.info.env.enabled=true",
```
onobc, line 45: Ah yeh, disabled since 2.6.3

applications/stream-applications-core/stream-applications-security-common/src/test/java/org/springframework/cloud/stream/app/security/common/ReactiveSecurityDisabledManagementSecurityEnabledTests.java
```diff
@@ -38,8 +37,8 @@
 		"spring.main.web-application-type=reactive",
 		"spring.cloud.streamapp.security.enabled=false",
 		"management.endpoints.web.exposure.include=health,info,bindings,env",
+		"management.info.env.enabled=true",
 		"info.name=MY TEST APP"})
```
onobc, line 41: For posterity, should we add a test that shows we can invoke the POST w/o creds when security is disabled?

dturanski: Good catch. I had the same thought.

applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebSecurityAutoConfiguration.java
dturanski: I believe that's the case once it gets here, but I'll add a check.

applications/stream-applications-core/stream-applications-security-common/src/main/java/org/springframework/cloud/stream/app/security/common/AppStarterWebSecurityAutoConfiguration.java
```diff
@@ -126,4 +127,38 @@ protected void configure(HttpSecurity http) throws Exception {
...
+		 * For example: <pre class="code">
+		 * EndpointRequest.to(HttpMethod.POST, "loggers")
+		 * </pre>
+		 * @param httpMethod the http method to include
+		 * @param endpoints the endpoints to include
+		 * @return the configured {@link RequestMatcher}
+		 */
+		static RequestMatcher to(HttpMethod httpMethod, String... endpoints) {
+			final EndpointRequest.EndpointRequestMatcher matcher = EndpointRequest.to(endpoints);
+			return (request) -> {
```
onobc, line 146: **[TinyNit]** we could make this a bit more DRY and create a private helper that takes in a request matcher guarded by httpmethod which would then be called by both of these. Not a big deal either way though.

applications/stream-applications-core/stream-applications-security-common/src/test/java/org/springframework/cloud/stream/app/security/common/SecurityEnabledAuthorizedAccessTests.java
```diff
@@ -73,6 +73,7 @@ public void testHealthEndpoint() {
 	@Test
 	@SuppressWarnings("rawtypes")
 	public void testInfoEndpoint() {
+		restTemplate.getRestTemplate().getInterceptors().clear();
```
onobc, line 76: So this test is really doing a "when security is enabled and I am logged in I can access X,Y,Z" so I think we should not clear out the interceptors.

onobc: This really brings up a good point (which is outside of the scope of this PR) that a 1-liner high level description  on each of these tests would go a long way as they are all very similiar looking and take a minute to get my head around which one is doing what.

onobc (changes requested):
I think we should leave the interceptors in the test https://github.com/spring-cloud/stream-applications/pull/212/files/bbfcb601d86901fd0f58519d758d0e40f8dcaff6..b84bd0df47901c551027d72ddc2ad14e634bf4d4#r792181914

sobychacko (approved):

dturanski:
merged.

PR #213, opened 2022-01-25
Re-enable and fix tests that were failing w/ Boot 2.6 upgrade.

onobc:
I will comment inline where appropriate. 

This fixes all previously disabled tests except the security ones which are being handled in @dturanski [code proposal](https://github.com/spring-cloud/stream-applications/pull/212).

Also, @artembilan I went ahead and kept the MongoDB tests in this PR. If you are ok w/ that we can close https://github.com/spring-cloud/stream-applications/pull/211.

### DIFF
--- a/applications/processor/script-processor/src/test/java/org/springframework/cloud/stream/app/processor/script/ScriptProcessorIntegrationTests.java
+++ b/applications/processor/script-processor/src/test/java/org/springframework/cloud/stream/app/processor/script/ScriptProcessorIntegrationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2021 the original author or authors.
+ * Copyright 2015-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,7 +19,6 @@ package org.springframework.cloud.stream.app.processor.script;
 import java.io.IOException;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.boot.WebApplicationType;
@@ -44,7 +43,6 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Chris Schaefer
  * @author Soby Chacko
  */
-@Disabled
 public class ScriptProcessorIntegrationTests {
 
 	@Test
--- a/applications/sink/redis-sink/pom.xml
+++ b/applications/sink/redis-sink/pom.xml
@@ -16,7 +16,7 @@
     </parent>
 
     <properties>
-        <embedded-redis.version>2.0.9</embedded-redis.version>
+        <embedded-redis.version>2.0.11</embedded-redis.version>
     </properties>
 
     <dependencies>
--- a/applications/sink/redis-sink/pom.xml
+++ b/applications/sink/redis-sink/pom.xml
@@ -16,7 +16,6 @@
     </parent>
 
     <properties>
-        <spring-cloud-starters.version>3.0.3</spring-cloud-starters.version>
         <embedded-redis.version>2.0.9</embedded-redis.version>
     </properties>
 
--- a/functions/consumer/tcp-consumer/src/main/java/org/springframework/cloud/fn/consumer/tcp/TcpConsumerConfiguration.java
+++ b/functions/consumer/tcp-consumer/src/main/java/org/springframework/cloud/fn/consumer/tcp/TcpConsumerConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 the original author or authors.
+ * Copyright 2015-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,7 +18,6 @@ package org.springframework.cloud.fn.consumer.tcp;
 
 import java.util.function.Consumer;
 
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.fn.common.tcp.EncoderDecoderFactoryBean;
@@ -38,28 +37,28 @@ import org.springframework.messaging.Message;
  *
  * @author Gary Russell
  * @author Christian Tzolov
+ * @author Chris Bono
  */
-@Configuration
+@Configuration(proxyBeanMethods = false)
 @EnableConfigurationProperties({TcpConsumerProperties.class, TcpConnectionFactoryProperties.class})
 public class TcpConsumerConfiguration {
 
-	@Autowired
 	private TcpConsumerProperties properties;
-
-	@Autowired
 	private TcpConnectionFactoryProperties tcpConnectionProperties;
 
-	@Qualifier("tcpSinkConnectionFactory")
-	@Autowired
-	private AbstractConnectionFactory connectionFactory;
-
-	@Bean
-	public Consumer<Message<?>> tcpConsumer() {
-		return handler()::handleMessage;
+	public TcpConsumerConfiguration(TcpConsumerProperties properties,
+									TcpConnectionFactoryProperties tcpConnectionProperties) {
+		this.properties = properties;
+		this.tcpConnectionProperties = tcpConnectionProperties;
 	}
 
 	@Bean
-	public TcpSendingMessageHandlerSmartLifeCycle handler() {
+	public Consumer<Message<?>> tcpConsumer(TcpSendingMessageHandlerSmartLifeCycle handler) {
+		return handler::handleMessage;
+	}
+
+	@Bean
+	public TcpSendingMessageHandlerSmartLifeCycle handler(@Qualifier("tcpSinkConnectionFactory") AbstractConnectionFactory connectionFactory) {
 		TcpSendingMessageHandlerSmartLifeCycle tcpMessageHandler = new TcpSendingMessageHandlerSmartLifeCycle();
 		tcpMessageHandler.setConnectionFactory(connectionFactory);
 		return tcpMessageHandler;
--- a/functions/consumer/tcp-consumer/src/test/java/org/springframework/cloud/fn/consumer/tcp/L4Tests.java
+++ b/functions/consumer/tcp-consumer/src/test/java/org/springframework/cloud/fn/consumer/tcp/L4Tests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 the original author or authors.
+ * Copyright 2015-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.fn.consumer.tcp;
 
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.integration.ip.tcp.serializer.ByteArrayLengthHeaderSerializer;
@@ -25,8 +24,7 @@ import org.springframework.test.context.TestPropertySource;
 /**
  * @author Gary Russell
  */
-@TestPropertySource(properties = { "tcp.consumer.encoder = L4" })
-@Disabled
+@TestPropertySource(properties = {"tcp.consumer.encoder = L4"})
 public class L4Tests extends AbstractTcpConsumerTests {
 
 	@Test
--- a/functions/consumer/websocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/websocket/WebsocketConsumerConfiguration.java
+++ b/functions/consumer/websocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/websocket/WebsocketConsumerConfiguration.java
@@ -62,24 +62,6 @@ public class WebsocketConsumerConfiguration {
 		websocketConsumerServer.run();
 	}
 
-	@Configuration
-	static class WebsocketConsumerServerConfiguration {
-		@Bean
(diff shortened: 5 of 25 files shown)


applications/sink/redis-sink/pom.xml
```diff
@@ -16,7 +16,6 @@
     </parent>
 
     <properties>
-        <spring-cloud-starters.version>3.0.3</spring-cloud-starters.version>
```
onobc, line 19: I bumped this to 3.1.1 and moved to `stream-applications-build`

functions/consumer/tcp-consumer/src/test/java/org/springframework/cloud/fn/consumer/tcp/L4Tests.java
```diff
@@ -25,8 +24,7 @@
 /**
  * @author Gary Russell
  */
-@TestPropertySource(properties = { "tcp.consumer.encoder = L4" })
-@Disabled
+@TestPropertySource(properties = {"tcp.consumer.encoder = L4"})
```
onobc, line 27: I did not intentionally reformat the brackets, the spring-cloud-build formatter did this.

functions/supplier/twitter-supplier/src/main/java/org/springframework/cloud/fn/supplier/twitter/status/search/TwitterSearchSupplierConfiguration.java
```diff
@@ -56,9 +56,6 @@
 	@Autowired
 	private Twitter twitter;
 
-	@Autowired
-	private SearchPagination searchPage;
```
onobc, line 60: This was causing circular reference as TwitterSearchSupplierConfiguration needs it's SearchPagination field injected, but wait, who will provide that to it? Ahh, the TwitterSearchSupplierConfiguration will supply it in https://github.com/spring-cloud/stream-applications/pull/213/files#diff-68c351c4c53dd25a5a570e9483de5205777402f7059498816f6a520bd58038dbR63. Ok, so let's create one of those TwitterSearchSupplierConfiguration , etc....

IOW, the configuration that creates the bean can not also inject it onto itself.

functions/consumer/websocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/websocket/WebsocketConsumerServer.java
```diff
@@ -33,34 +33,36 @@
...
  * @author Gary Russell
+ * @author Chris Bono
  */
 public class WebsocketConsumerServer {
 
 	private static final Log logger = LogFactory.getLog(WebsocketConsumerServer.class);
 
 	static final List<Channel> channels = Collections.synchronizedList(new ArrayList<Channel>());
 
-	@Autowired
```
onobc, line 52: This was working before via indirect field injection as part of ...

```java
@Bean
public WebsocketConsumerServer server() {
	return new WebsocketConsumerServer();
}
```
This style of field injection is not considered a best practice and is harder to follow how the fields get populated (unless you know about it). Constructor injection is preferred in this case.

functions/consumer/websocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/websocket/WebsocketConsumerConfiguration.java
```diff
@@ -42,42 +43,51 @@
...
-
 	private static final Log logger = LogFactory.getLog(WebsocketConsumerConfiguration.class);
 
-	private final InMemoryTraceRepository websocketTraceRepository = new InMemoryTraceRepository();
-
 	@Value("${endpoints.websocketconsumertrace.enabled:false}")
 	private boolean traceEndpointEnabled;
 
+	@Autowired
+	private WebsocketConsumerServer websocketConsumerServer;
```
onobc, line 58: This was another case of needing the configuration to create the thing that it needs to be injected w/. In this case it was the `server()` bean call in the `init` method. 

To get around this, I switched to wrapping the server aspects in a nested static configuration class so that those beans will be created by that configuration and then ultimately injected into the outer configuration.

functions/consumer/tcp-consumer/src/main/java/org/springframework/cloud/fn/consumer/tcp/TcpConsumerConfiguration.java
```diff
@@ -38,28 +37,28 @@
...
-	private AbstractConnectionFactory connectionFactory;
+	public TcpConsumerConfiguration(TcpConsumerProperties properties,
+									TcpConnectionFactoryProperties tcpConnectionProperties) {
+		this.properties = properties;
+		this.tcpConnectionProperties = tcpConnectionProperties;
+	}
 
 	@Bean
-	public Consumer<Message<?>> tcpConsumer() {
-		return handler()::handleMessage;
```
onobc, line 58: Again, the config trying to use the thing that it creates via field injection. I moved this to constructor injection and bean method param injection.

functions/consumer/tcp-consumer/src/main/java/org/springframework/cloud/fn/consumer/tcp/TcpConsumerConfiguration.java
```diff
@@ -38,28 +37,28 @@
...
-	@Autowired
 	private TcpConsumerProperties properties;
-
-	@Autowired
 	private TcpConnectionFactoryProperties tcpConnectionProperties;
 
-	@Qualifier("tcpSinkConnectionFactory")
-	@Autowired
-	private AbstractConnectionFactory connectionFactory;
+	public TcpConsumerConfiguration(TcpConsumerProperties properties,
```
onobc, line 49: Do we consider these configurations as public API? This change (new non-default constructor) removes the default constructor. If someone was constructing one of these manually for some reason, it would not compile anymore. 

Since this is a minor release this should be acceptable - just wanted to call it out though.

artembilan (changes requested):
Thank you, @onobc , for outstanding work, but I think this is too complicated.
You have raised a lot of concerns, which probably would be better to discuss individually. 

I think for now it will be OK to review and merge this as one. Although it would take some time for us to digest all the changes and for you to bring fixes for our requests.

I'll check your suggestion for MongoDB, but one of the explicit version didn't work for me on Windows. Then I'll apply that change into my PR and you can merge it and rebase yours. So, your one would become smaller 😄 .

Stay tuned for some review later today

functions/consumer/websocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/websocket/WebsocketConsumerServer.java
```diff
@@ -33,34 +33,36 @@
...
-	WebsocketConsumerServerInitializer initializer;
+	private WebsocketConsumerServerInitializer initializer;
 
 	private EventLoopGroup bossGroup;
 
 	private EventLoopGroup workerGroup;
 
 	private int port;
 
+	public WebsocketConsumerServer(WebsocketConsumerProperties properties, WebsocketConsumerServerInitializer initializer) {
```
onobc, line 61: [Ditto](https://github.com/spring-cloud/stream-applications/pull/213/files#r791765082)...

functions/supplier/twitter-supplier/src/main/java/org/springframework/cloud/fn/supplier/twitter/status/search/TwitterSearchSupplierConfiguration.java
```diff
@@ -71,18 +68,18 @@ public SearchPagination searchPage() {
 
 
 	@Bean
-	public Supplier<Message<byte[]>> twitterSearchSupplier() {
+	public Supplier<Message<byte[]>> twitterSearchSupplier(SearchPagination searchPage) {
```
onobc, line 71: [Ditto](https://github.com/spring-cloud/stream-applications/pull/213/files#r791765082)...

applications/processor/script-processor/src/test/java/org/springframework/cloud/stream/app/processor/script/ScriptProcessorIntegrationTests.java
```diff
@@ -44,7 +43,6 @@
  * @author Chris Schaefer
  * @author Soby Chacko
  */
-@Disabled
```
artembilan, line 47: @sobychacko , what was wrong with this test?
Doesn't look like it fails somehow: I really just removed this annotation and it passes.

Thanks

applications/sink/redis-sink/pom.xml
```diff
@@ -16,7 +16,6 @@
...
         <embedded-redis.version>2.0.9</embedded-redis.version>
```
artembilan, line 19: There is a newer version for this: https://mvnrepository.com/artifact/com.playtika.testcontainers/embedded-redis.

functions/consumer/tcp-consumer/src/main/java/org/springframework/cloud/fn/consumer/tcp/TcpConsumerConfiguration.java
```diff
@@ -38,28 +37,28 @@
  *
  * @author Gary Russell
  * @author Christian Tzolov
+ * @author Chris Bono
  */
-@Configuration
+@Configuration(proxyBeanMethods = false)
```
artembilan, line 42: This is what I asked long time ago to revise...

Thank you, @onobc , for starting this effort!

functions/consumer/tcp-consumer/src/main/java/org/springframework/cloud/fn/consumer/tcp/TcpConsumerConfiguration.java
artembilan: I don't think the field injection is too evil, but for consistency it really won't be bad to move autowiring via ctor.
This way when end-user would extend this class, he/she would have access to these injected beans in the overridden ctor.

functions/consumer/websocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/websocket/WebsocketConsumerServer.java
artembilan: That's correct. It might be OK for end-user project. And I would say for this type of configuration I'd prefer a `@Component` and their scanning instead of `@Bean`.
But when we develop a library it is better to stay away from autowiring at all.

stream-applications-build/pom.xml
```diff
@@ -44,8 +44,9 @@
 		<spring-kafka.version>2.8.2</spring-kafka.version>
 		<spring-rabbit.version>2.4.2</spring-rabbit.version>
 		<spring-cloud-function.version>3.2.1</spring-cloud-function.version>
+		<spring-cloud-starters.version>3.1.0</spring-cloud-starters.version>
 		<spring-data-geode-test.version>0.0.21.RELEASE</spring-data-geode-test.version>
-		<test-containers.version>1.15.2</test-containers.version>
+		<test-containers.version>1.15.3</test-containers.version>
```
artembilan, line 49: ??? According Maven we have already `1.16.3`: https://mvnrepository.com/artifact/org.testcontainers/testcontainers. 🤷

artembilan (changes requested):
Some review.
Thanks

applications/processor/script-processor/src/test/java/org/springframework/cloud/stream/app/processor/script/ScriptProcessorIntegrationTests.java
onobc: It failed for me and I believe it was related to compatibility matrix (which was old and requiring boot 2.4,2.5). I would be sure to re-run `./setup-build.sh` to make sure it is using the versions in the poms. I definitely saw all of these tests fail before I "fixed" them.

functions/consumer/tcp-consumer/src/main/java/org/springframework/cloud/fn/consumer/tcp/TcpConsumerConfiguration.java
onobc: No worries! Yes, I have tried to do this in every single `@Configuration` I have touched since Boot 2.2 (or whenever it was added).

stream-applications-build/pom.xml
onobc: Got it. I will make sure this is at the proper version. I thought that was 1.15.3 but will 2x check what SpringBoot is also recommending in their deps.

artembilan: This is test dependency, so as long as it does not require `main` changes, we are good to go.

onobc:
> Thank you, @onobc , for outstanding work, but I think this is too complicated. You have raised a lot of concerns, which probably would be better to discuss individually.
> 
> I think for now it will be OK to review and merge this as one. Although it would take some time for us to digest all the changes and for you to bring fixes for our requests.
> 
> I'll check your suggestion for MongoDB, but one of the explicit version didn't work for me on Windows. Then I'll apply that change into my PR and you can merge it and rebase yours. So, your one would become smaller 😄 .
> 
> Stay tuned for some review later today

I too am not a fan of large PRs and I could have done a better job of breaking this down into digestable chunks. I instead went in, heads down and got them all working at once. 

I think it looks bigger/worse than it is. The 3 config class changes are breaking circular deps by moving to ctro injection. 

I would be happy to Zoom as well and we can walk through that way if you like. Just lmk.

applications/sink/redis-sink/pom.xml
onobc: I think we should stick on [2.0.11](https://github.com/Playtika/testcontainers-spring-boot/tree/2.0.11/embedded-redis) since that is the last version before they jump to Redis 6.2 for the docker image and Spring Data Redis is tested against the latest [4.0 release](https://docs.spring.io/spring-data/redis/docs/current/reference/html/#requirements).

I will bump to 2.0.11

artembilan (approved):

PR #217, opened 2022-01-27
Break dependence on RabbitAutoConfiguration

onobc:
This code proposal breaks the dependence on [RabbitAutoConfiguration](https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/187#issuecomment-860227374) for the `rabbit-consumer` and `rabbit-supplier` functions.

### DIFF
--- a/applications/source/rabbit-source/src/test/java/org/springframework/cloud/stream/app/source/rabbit/RabbitSourceListenerTests.java
+++ b/applications/source/rabbit-source/src/test/java/org/springframework/cloud/stream/app/source/rabbit/RabbitSourceListenerTests.java
@@ -107,11 +107,12 @@ public class RabbitSourceListenerTests {
 								"${spring.rabbitmq.test.port}"
 				)) {
 
-			final RabbitTemplate rabbitTemplate = context.getBean(RabbitTemplate.class);
+			// Reset the boot connection factory -should not matter to container as it SHOULD be using its own connection factory
 			final CachingConnectionFactory bootFactory = context.getBean(CachingConnectionFactory.class);
-			rabbitTemplate.convertAndSend("scsapp-testOwnSource", "foo");
-
 			bootFactory.resetConnection();
+
+			// Send a message on a separate connection - the container should still receive it.
+			sendMessageOnSeparateConnection("scsapp-testOwnSource", "foo", bootFactory);
 
 			OutputDestination target = context.getBean(OutputDestination.class);
 			Message<byte[]> sourceMessage = target.receive(600000, "rabbitSupplier-out-0");
@@ -119,6 +120,32 @@ public class RabbitSourceListenerTests {
 			final String actual = new String(sourceMessage.getPayload());
 			assertThat(actual).isEqualTo("foo");
 			assertThat(bootFactory.getCacheProperties().getProperty("localPort")).isEqualTo("0");
+		}
+	}
+
+	/**
+	 * Sends a message on a separate connection.
+	 *
+	 * @param routingKey message routing key
+	 * @param payload message content
+	 * @param bootFactory the auto-configured connection factory used to get connection coordinates from
+	 */
+	private void sendMessageOnSeparateConnection(String routingKey, Object payload, CachingConnectionFactory bootFactory) {
+		CachingConnectionFactory copiedConnectionFactory = null;
+		try {
+			copiedConnectionFactory = new CachingConnectionFactory(bootFactory.getHost(), bootFactory.getPort());
+			copiedConnectionFactory.setUsername(bootFactory.getUsername());
+			copiedConnectionFactory.setPassword(bootFactory.getRabbitConnectionFactory().getPassword());
+			if (bootFactory.getVirtualHost() != null) {
+				copiedConnectionFactory.setVirtualHost(bootFactory.getVirtualHost());
+			}
+			RabbitTemplate rabbitTemplate = new RabbitTemplate(copiedConnectionFactory);
+			rabbitTemplate.convertAndSend(routingKey, payload);
+		}
+		finally {
+			if (copiedConnectionFactory != null) {
+				copiedConnectionFactory.resetConnection();
+			}
 		}
 	}
 
--- a/functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
+++ b/functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
@@ -24,7 +24,7 @@ import com.rabbitmq.client.impl.CredentialsRefreshService;
 import org.springframework.amqp.core.MessageDeliveryMode;
 import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
-import org.springframework.amqp.rabbit.connection.ConnectionNameStrategy;
+import org.springframework.amqp.rabbit.connection.RabbitConnectionFactoryBean;
 import org.springframework.amqp.rabbit.core.RabbitTemplate;
 import org.springframework.amqp.support.converter.Jackson2JsonMessageConverter;
 import org.springframework.amqp.support.converter.MessageConverter;
@@ -33,8 +33,9 @@ import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.amqp.CachingConnectionFactoryConfigurer;
 import org.springframework.boot.autoconfigure.amqp.ConnectionFactoryCustomizer;
-import org.springframework.boot.autoconfigure.amqp.RabbitAutoConfiguration;
+import org.springframework.boot.autoconfigure.amqp.RabbitConnectionFactoryBeanConfigurer;
 import org.springframework.boot.autoconfigure.amqp.RabbitProperties;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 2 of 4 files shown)


onobc:
@artembilan @sobychacko 

I closed the original PR and created this fresh one (sans boot26 hacks and docs that were in the other one).

Also, note that I broke this into 2 commits. The first shows that the state of the own-connection feature was in fact busted currently. The rabbit consumer test was disabled and the rabbit supplier test was not accurate (it is now fixed). If you cherrypick the 1st commit onto main locally you can see them both fail. 

The 2nd commit fixes this w/ the removal of the rabbitautoconfig dependence (which also restores the own-connection feature)

applications/source/rabbit-source/src/test/java/org/springframework/cloud/stream/app/source/rabbit/RabbitSourceListenerTests.java
```diff
@@ -122,6 +123,31 @@ public void testOwnConnection() {
...
 	}
 
+	/**
+	 * Sends a message on a separate connection.
+	 *
+	 * @param routingKey message routing key
+	 * @param payload message content
+	 * @param bootFactory the auto-configured connection factory used to get connection coordinates from
+	 */
+	private void sendMessageOnSeparateConnection(String routingKey, Object payload, CachingConnectionFactory bootFactory) {
```
onobc, line 133: Yeh, its gross. We can't however create another RabbitTemplate nor CCF as that will disable SB auto-config.

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
```diff
@@ -47,16 +48,21 @@
...
+ */
 @EnableConfigurationProperties(RabbitConsumerProperties.class)
 @Configuration
 public class RabbitConsumerConfiguration implements DisposableBean {
 
 	@Autowired
 	private RabbitProperties bootProperties;
 
-	@Autowired
-	private ObjectProvider<ConnectionNameStrategy> connectionNameStrategy;
```
onobc, line 58: This was only in here to appease the signature of the connection factory creator. The code immediately turned arounds and overrides the naming strategy so I removed this.

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
```diff
@@ -47,16 +48,21 @@
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHandler;
 
+/**
+ * A configuration for RabbitMQ Consumer function. Uses a
+ * {@link AmqpOutboundChannelAdapterSpec} to save payload contents to RabbitMQ.
+ *
+ * @author Soby Chako
+ * @author Nicolas Labrot
```
onobc, line 56: I reverse engineered this author list from git history.

functions/supplier/rabbit-supplier/src/main/java/org/springframework/cloud/fn/supplier/rabbit/RabbitSupplierConfiguration.java
```diff
@@ -77,9 +79,6 @@ public MessageProperties toMessageProperties(AMQP.BasicProperties source,
 	@Autowired
 	private RabbitProperties rabbitProperties;
 
-	@Autowired
```
onobc, line 80: [Ditto...](https://github.com/spring-cloud/stream-applications/pull/217/files#r793213104)

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
```diff
@@ -144,25 +146,49 @@ public void destroy() {
...
+	 * @throws Exception
+	 */
+	private CachingConnectionFactory rabbitConnectionFactory(RabbitProperties properties, ResourceLoader resourceLoader,
+			ObjectProvider<CredentialsProvider> credentialsProvider,
+			ObjectProvider<CredentialsRefreshService> credentialsRefreshService,
+			ObjectProvider<ConnectionFactoryCustomizer> connectionFactoryCustomizers) throws Exception {
+
+		/* NOTE: This is based on RabbitAutoConfiguration.RabbitConnectionFactoryCreator
+		 * 		https://github.com/spring-projects/spring-boot/blob/c820ad01a108d419d8548265b8a34ed7c5591f7c/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java#L95
+		 * [UPGRADE_CONSIDERATION] this should stay somewhat in sync w/ the functionality provided by its original source.
```
onobc, line 172: I have used tokens in the source code like this in the past to allow an easy search across code when upgrading SB when I have modeled some configuration closely to what is in the auto-config. 

Wdyt about this @sobychacko @artembilan ?

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
```diff
@@ -144,25 +146,49 @@ public void destroy() {
...
+		 */
+		RabbitConnectionFactoryBean connectionFactoryBean = new RabbitConnectionFactoryBean();
```
onobc, line 174: Because this code is identical in the supplier and consumer I was thinking we could move it somewhere common and share it between the 2.

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
```diff
@@ -144,25 +146,49 @@ public void destroy() {
...
+		CachingConnectionFactory cachingConnectionFactory = new CachingConnectionFactory(connectionFactory);
+		CachingConnectionFactoryConfigurer cachingConnectionFactoryConfigurer = new CachingConnectionFactoryConfigurer(properties);
+		cachingConnectionFactoryConfigurer.setConnectionNameStrategy(cf -> "rabbit.sink.own.connection");
+		cachingConnectionFactoryConfigurer.configure(cachingConnectionFactory);
+		cachingConnectionFactory.afterPropertiesSet();
+
+		return cachingConnectionFactory;
+	}
 
+}
```
artembilan, line 194: New line in the end of each file.
But still: I guess the formatter should do that for us.
Only need to learn how to use it 😄

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
```diff
@@ -144,25 +146,49 @@ public void destroy() {
...
+
+		com.rabbitmq.client.ConnectionFactory connectionFactory = connectionFactoryBean.getObject();
+		connectionFactoryCustomizers.orderedStream()
+				.forEach((customizer) -> customizer.customize(connectionFactory));
+
+		CachingConnectionFactory cachingConnectionFactory = new CachingConnectionFactory(connectionFactory);
+		CachingConnectionFactoryConfigurer cachingConnectionFactoryConfigurer = new CachingConnectionFactoryConfigurer(properties);
+		cachingConnectionFactoryConfigurer.setConnectionNameStrategy(cf -> "rabbit.sink.own.connection");
+		cachingConnectionFactoryConfigurer.configure(cachingConnectionFactory);
+		cachingConnectionFactory.afterPropertiesSet();
```
artembilan, line 189: All of these manual `afterPropertiesSet()` scare me since they are done without an `ApplicationContext` propagation and during other bean definition.
Although it might be OK for us so far since we just need a lite `CF` instance not tied with `AC`.
If you confirm that tests work well with such an approach, then we won't spend too much time on this and just move on.

artembilan (changes requested):
Just a couple easy comments.
If tests are OK, and Checkstyle doesn't complain, we are good to merge this.

Thank you!

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
onobc: I agree. Is javaformat plugin installed in this repo? Or are you saying we should add it?

artembilan: No, I only see `spring-javaformat-checkstyle` which does don't do reformatting: https://github.com/spring-io/spring-javaformat#source-formatting
Sorry for a noise.

We may reconsider this in a different PR

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
artembilan: Fine with me

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
onobc: I reformatted w/in IDEA (using the spring cloud build formatter) and it did correct the finally. I am about to submit another commit w/ that as well as a checkstyle fix.

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
artembilan: Right. For other similar we have a dedicated module in the `common`, e.g. `mqtt-common`, `aws-s3-common` etc.

We may have `rabbit-common` for this "common" code, but does it really justify already a whole module?

(Let's revise it separately!)

functions/consumer/rabbit-consumer/src/main/java/org/springframework/cloud/fn/consumer/rabbit/RabbitConsumerConfiguration.java
onobc: This is taken from what SB [RabbitAutoConfiguration](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java#L127) does. I think its fine because the impl of CF has what it needs to create the instance. But I hear ya, I typically raise eyebrow when I see this too.

artembilan: Well, a bit different though: that `RabbitConnectionFactoryBean` is fully OK. I worry more about this `CachingConnectionFactory` which comes with contract like this:
```
implements ConnectionFactory, DisposableBean, BeanNameAware,
		ApplicationContextAware, ApplicationEventPublisherAware, ApplicationListener<ContextClosedEvent>
```

But again: if we are on the same page and there is no harm that we don't tie it with the `AC`, then it is OK to go without those hooks and callback.

sobychacko (approved):

PR #218, opened 2022-01-27
Add requiresUnpack for script-processor

artembilan:
The `Jython` doesn't work in the uber jar as a nested one.
The Spring Boot provides a trick via unpacking requested nested jars.

* Use the latest `spring-cloud-dataflow-apps-generator-plugin` which allows now
to propagate a custom configuration for `spring-boot-maven-plugin`
* Add `requiresUnpack` for `jython-standalone` into the `script-processor`

**Cherry-pick to 2021.1.x**

### DIFF
--- a/applications/processor/script-processor/pom.xml
+++ b/applications/processor/script-processor/pom.xml
@@ -121,6 +121,18 @@
                                 </dependency>
                             </dependencies>
                         </maven>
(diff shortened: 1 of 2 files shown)


sobychacko:
Looks good. Merged upstream.

PR #219, opened 2022-01-28
Fix rabbit sink IT OwnConnectionTest

onobc:
* Check boot connection factory destroyed earlier in test
* Check rabbit template not using boot connection factory

Explanation inline  in PR

### DIFF
--- a/applications/sink/rabbit-sink/src/test/java/org/springframework/cloud/stream/app/sink/rabbit/OwnConnectionTest.java
+++ b/applications/sink/rabbit-sink/src/test/java/org/springframework/cloud/stream/app/sink/rabbit/OwnConnectionTest.java
@@ -25,6 +25,10 @@ import org.springframework.test.context.TestPropertySource;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+/**
+ * @author Soby Chacko
+ * @author Chris Bono
+ */
 @TestPropertySource(properties = {"rabbit.routingKey=scsapp-testOwn",
 		"rabbit.own-connection=true"})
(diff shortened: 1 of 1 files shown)


applications/sink/rabbit-sink/src/test/java/org/springframework/cloud/stream/app/sink/rabbit/OwnConnectionTest.java
```diff
@@ -33,12 +37,18 @@
 	public void test() {
 		this.rabbitAdmin.declareQueue(
 				new Queue("scsapp-testOwn", false, false, true));
-		this.bootFactory.resetConnection();
```
onobc, line 36: This was some timing related issue due to the fact that the RabbitAdmin uses the boot conn factory and if it attempts to do any operations it will recreate a connection and then the "check that localPort == 0" will no longer be true. 

I am not sure what the admin was trying to do when the test failed as I was unable to consistently reproduce it - I reproduced it maybe 3 times out of 50. 

### Here is what we are trying to test:
"Since the outbound adapter uses its own CF, we should be able to destroy/make-not-usable the boot created CF and all should still be well". 

### Here is what we WERE using to verify the above: ​
- "Reset boot CF"
- "Send/receive message - aka all is well"
- "Check boot CF has localPort=0"

We are not guaranteed that the boot CF will stay w/ localPort=0.

### Here is what we ARE now using to verify the above:
- "Destroy boot CF"
- "Check boot CF has localPort=0"=
- "Make sure rabbit template not using boot CF"
- "Send/receive mesage - aka all is well"

sobychacko:
Looks good. Merged upstream.

PR #228, opened 2022-02-02
Set spring.integration.jdbc.platform in jdbc-sink app

onobc:
Fixes IT failure KafkaJdbcLogStreamTests

The `jdbc-source-kafka` is failing to properly start. It is attempting to determine the DB platform type from an actual connection but the DB is not yet up. This is due to the rework of DB init scripts in 2.6 ([bug outlined here](https://github.com/spring-projects/spring-boot/issues/28897)).

~This sets `spring.integration.jdbc.platform=mysql` to specify the platform.~

This now uses `@Nested` to wait for the outer `@Container` to start.

### DIFF
--- a/applications/source/jdbc-source/src/main/resources/application.properties
+++ b/applications/source/jdbc-source/src/main/resources/application.properties
@@ -1,0 +1,1 @@
+spring.integration.jdbc.platform=mysql
--- a/applications/source/jdbc-source/src/main/resources/application.properties
+++ b/applications/source/jdbc-source/src/main/resources/application.properties
@@ -1,1 +1,0 @@
-spring.integration.jdbc.platform=mysql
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
> the DB is not yet up

Doesn't that mean that our `streamApp` container must wait for `mySQL` one in our tests?

Isn't that something like reconciliation should happen in case of container dependencies?

onobc:
> > the DB is not yet up
> 
> Doesn't that mean that our `streamApp` container must wait for `mySQL` one in our tests?
> 
> Isn't that something like reconciliation should happen in case of container dependencies?

I thought of this as well but I opted for this simpler fix because:
1. I was not sure how to do that w/ the current test layout. 
2. I want a green CI run :) 
3. Also, this is the same way it has been the whole time (they start up closely to one another and the server code does wait for the message so it is ok). 

However, the way that I have implemented it would force `mysql` on all `jdbc-source` generated apps. 

I think you are correct in that we should probably wait to start the app container until the mysql container is up and remove this property.

PR #244, opened 2022-03-02
Quiet down integration test logging

dturanski:
Add logback.xml to set TestContainers logging level.

Add spring-boot-geode dependency management for geode apps, required for `spring-geode-starter-logging` which fixes the logging config.


applications/sink/geode-sink/src/test/resources/logback-include.xml
```diff
@@ -0,0 +1,16 @@
...
+		</encoder>
+	</appender>
+
+	<appender name="delegate" class="org.springframework.geode.logging.slf4j.logback.DelegatingAppender"/>
+
+	<logger name="com.gemstone.gemfire" level="${spring.boot.data.gemfire.log.level:-WARN}"/>
+	<logger name="org.apache.geode" level="${spring.boot.data.gemfire.log.level:-WARN}"/>
+	<logger name="org.jgroups" level="${spring.boot.data.gemfire.jgroups.log.level:-ERROR}"/>
+
+</included>
```
onobc, line 16: All of the logback-include.xml have no newline at end. Github lighting that up.

applications/sink/geode-sink/src/test/resources/logback-include.xml
```diff
@@ -0,0 +1,16 @@
...
+
+	<appender name="console" class="ch.qos.logback.core.ConsoleAppender">
+		<encoder>
+			<pattern>%d %5p %40.40c:%4L - %m%n</pattern>
+		</encoder>
+	</appender>
+
+	<appender name="delegate" class="org.springframework.geode.logging.slf4j.logback.DelegatingAppender"/>
+
+	<logger name="com.gemstone.gemfire" level="${spring.boot.data.gemfire.log.level:-WARN}"/>
```
onobc, line 12: Is the property key something specific to the Geode starter or did we create it? We could instead just set a default level and then recommend the standard [Spring Boot method](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.logging.log-levels) of adjusting the log level settings.

onobc (commented):
@dturanski LGTM - thanks for doing this. 

I left a couple comments for you.

applications/sink/geode-sink/src/test/resources/logback-include.xml
dturanski: This is probably cut and paste from an example somewhere.  The geode apps introduced a dependency on `spring-boot-data-gemfire` which came subsequent to these apps. Geode uses proprietary logging config, so it's got its own property now.  I noticed the pattern is different than the others. I don't have a strong opinion on this.

applications/sink/mqtt-sink/src/test/resources/logback.xml
```diff
@@ -0,0 +1,13 @@
+<configuration>
+
+	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+		<encoder>
+			<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
```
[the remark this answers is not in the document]
dturanski: They are the same now

dturanski:
OK. Reverted the timesource metadata changes. Should be good to go.

onobc (approved):

onobc:
Closed via https://github.com/spring-cloud/stream-applications/commit/b8cb30fa9ea8d8b9a5ce2ec9e848fb41b243291c

Issue #248, opened 2022-03-24
aggregator-processor (3.2.0) does not start up


making:
```
java -jar aggregator-processor-kafka-3.2.0.jar ....
```

does not start and eats up the cpu. nothing output

```
docker run --rm springcloudstream/aggregator-processor-kafka:3.2.0
```

is the same.

I managed to capture the error log on k8s platform although it eats up the cpu as well

```
demo-aggr-aggregator-v1-8455d64597-qwspz demo-aggr-aggregator-v1 16:09:51.347 [main] ERROR org.springframework.boot.SpringApplication - Application run failed
demo-aggr-aggregator-v1-8455d64597-qwspz demo-aggr-aggregator-v1 java.lang.IllegalArgumentException: [/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/proc/self/task/1/cwd/sys/kernel/slab/shmem_inode_cache/cgroup/shmem_inode_cache(1450:cloud-final.service)] is not a valid directory
demo-aggr-aggregator-v1-8455d64597-qwspz demo-aggr-aggregator-v1 	at org.springframework.util.Assert.isTrue(Assert.java:121)
demo-aggr-aggregator-v1-8455d64597-qwspz demo-aggr-aggregator-v1 	at org.springframework.geode.boot.autoconfigure.SslAutoConfiguration.assertDirectory(SslAutoConfiguration.java:248)
demo-aggr-aggregator-v1-8455d64597-qwspz demo-aggr-aggregator-v1 	at org.springframework.geode.boot.autoconfigure.SslAutoConfiguration.locateKeyStoreInFileSystem(SslAutoConfiguration.java:213)
demo-aggr-aggregator-v1-8455d64597-qwspz demo-aggr-aggregator-v1 	at org.springframework.geode.boot.autoconfigure.SslAutoConfiguration.locateKeyStoreInFileSystem(SslAutoConfiguration.java:219)
[...]

PR #250, opened 2022-03-28
GH-248: Exclude Geode SSL config

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/248

The Geode SSL auto-config is done via `SslAutoConfiguration.SslEnvironmentPostProcessor`
which cannot be excluded as an auto-config.

* Set `SslAutoConfiguration.SECURITY_SSL_ENVIRONMENT_POST_PROCESSOR_ENABLED_PROPERTY`
to `false` from the `ExcludeStoresAutoConfigurationEnvironmentPostProcessor`
when the store type is not Geode


onobc:
I'll get to this one shortly.

onobc:
Closed via https://github.com/spring-cloud/stream-applications/commit/5d83512da4139d025103968ded9ec641bda9a6b0

Issue #249, opened 2022-03-24
Support metadata store in file source


making:
Currently the metadata store is supported by remote files source (S3, ftp and sftp), but not by file source.
There are cases where it is convenient to be able to use the metadata store with file source as well.

In my use case of SCDF on AKS, the Persistent Volume is mounted on the directory of the file source. The entity of this Persistent Volume is Azure File. Therefore, it is just a file source, but it is used as an “azure file source” like s3 source. Requires metadata support when restarted.
It will also be available as a “nfs source” or “vsan file service source” etc.

PR #251, opened 2022-03-28
GH-249: Add metadata-store to file-supplier

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/249

Currently, the metadata store is supported by remote files source (S3, ftp and sftp), but not by file source.
There are cases where it is convenient to be able to use the metadata store with file source as well.

* Add externally configured `ConcurrentMetadataStore` into `FileSupplierConfiguration`
which is based on the auto-configuration provided by the `metadata-store-common` artifact
* Test with an embedded JDBC store
* Remove redundant dependencies in the file-supplier pom - supplied by the parent
* Add `MetadataStoreProperties` to allow-list for dataflow
* Fix typo in the `s3-supplier` README


dturanski (approved):
LGTM;

PR #257, opened 2022-04-13
Update config metadata for polling sources

### LINKED ISSUE #247: Consider adding back in Poller properties for Source apps
The `DefaultPollerProperties` is gone from Spring Cloud Stream. This removed the `trigger.*` properties from the source apps. We should consider adding them back in by updating the `dataflow.configuration-metadata.properties` to
```
configuration-properties.classes=org.springframework.cloud.fn.supplier.time.TimeSupplierProperties,\
  org.springframework.boot.autoconfigure.integration.IntegrationProperties$Poller
```

> ⚠️ Doing this will (and should) break the recently adjusted SCDF tests. 

**Background**: https://vmware.slack.com/archives/C8XRX5NNM/p1646943272081199

dturanski:
Fixes #247

### DIFF
--- a/applications/source/jdbc-source/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
+++ b/applications/source/jdbc-source/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
@@ -1,5 +1,5 @@
 configuration-properties.classes=org.springframework.cloud.fn.supplier.jdbc.JdbcSupplierProperties,\
-  org.springframework.cloud.stream.config.DefaultPollerProperties
+  org.springframework.boot.autoconfigure.integration.IntegrationProperties$Poller
 configuration-properties.names=\
 spring.datasource.url,\
 spring.datasource.driver-class-name,\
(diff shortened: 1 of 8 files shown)


onobc (approved):
Looks good @dturanski . 

I do see that the apps README.md says at the bottom...

> Also see the [Spring Boot Documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html) for addition DataSource properties and TriggerProperties and MaxMessagesProperties for polling options.

I don't think it links directly to the trigger properties but I doubt they will find the integration properties in there. should we update this?

artembilan:
@onobc ,

both of those `TriggerProperties` and `MaxMessagesProperties` are covered now in that `IntegrationProperties.Poller`.
So, we probably just need to remove their mentioning everywhere from now on.

@dturanski ,

Don't we need to update all the READMEs with these metadata changes?
I guess we have some Maven task to do that automatically...

onobc:
> Don't we need to update all the READMEs with these metadata changes?

@artembilan that is what I was referring to. There is no mention of the poller properties currently in the READMEs - only a link to the boot docs. I think it would be helpful to add the config props to the README. The generator creates a README in each app but it just says "An XYZ stream app binder" (roughly). There is an actual hardcoded README in the app [root dir](https://github.com/spring-cloud/stream-applications/blob/main/applications/source/jdbc-source/README.adoc) that hardcode lists the properties AFAICT. We would need to update each README if we want to show these props.

onobc:
I verified that running `mvnw verify` in each app dir does update the readme. While this is done as part of the release build, I will do it manually in each of the above affected apps to get the changes into this PR.

onobc:
Closed via https://github.com/spring-cloud/stream-applications/commit/d0781606a139f2581e0cbea5bc41744f997b5ee0

PR #261, opened 2022-05-03
Update baseimage version

### LINKED ISSUE #259: Update the OpenJDK image
Investigate what needs to be updated [here](https://github.com/spring-cloud/baseimage/blame/main/Dockerfile). 

Ensure that the proper versions are aligned and used in the new apps.

corneil:
Fixes #259


build-images.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
onobc, line 1: Is this for debugging? Should we check it in?

applications/stream-applications-core/pom.xml
```diff
@@ -19,13 +19,13 @@
...
         <java-functions.version>1.2.0</java-functions.version>
-        <apps.base-image>springcloud/baseimage:1.0.0</apps.base-image>
+        <apps.base-image>springcloud/baseimage:1.0.1</apps.base-image>
         <mockserver.version>5.10</mockserver.version>
         <spring-cloud-stream-dependencies.version>3.2.2</spring-cloud-stream-dependencies.version>
         <spring-cloud-stream.version>3.2.2</spring-cloud-stream.version>
-        <spring-cloud-dataflow-apps-generator-plugin.version>1.0.6</spring-cloud-dataflow-apps-generator-plugin.version>
-        <spring-cloud-dataflow-apps-docs-plugin.version>1.0.6</spring-cloud-dataflow-apps-docs-plugin.version>
-        <spring-cloud-dataflow-apps-metadata-plugin.version>1.0.6</spring-cloud-dataflow-apps-metadata-plugin.version>
+        <spring-cloud-dataflow-apps-generator-plugin.version>1.0.7-SNAPSHOT</spring-cloud-dataflow-apps-generator-plugin.version>
```
onobc, line 26: This change is unrelated, correct?

corneil: This change is needed to ensure that the new version of app plugin is used so that baseimage version can be specified in stream-applications-core.

build-images.sh
corneil: It is useful when doing local testing. It isn't required by the CI build.

onobc: I am ok w/ adding this - do you mind adding comment to that effect at the top though?

PR #274, opened 2022-05-22
Update to Spring Boot 2.7.0

### LINKED ISSUE #272: Update Boot 2.6.8

onobc:
Fixes #272


functions/function/twitter-function/src/test/java/org/springframework/cloud/fn/twitter/users/TwitterUsersFunctionTests.java
```diff
@@ -165,7 +164,7 @@ public void testOne() throws IOException {
 
 	@TestPropertySource(properties = {
 			"twitter.users.type=lookup",
-			"twitter.users.lookup.screenName=#jsonPath(payload,'$..[*].code')"
+			"twitter.users.lookup.screenName=#jsonPath(payload,'$[*].code')"
```
onobc, line 167: Interesting enough, the latest libs (I think its JsonPath internally doing this) does not return results when the "deep scan - '..'" is added to the query. Works fine w/o it.

stream-applications-build/pom.xml
```diff
@@ -49,6 +49,7 @@
 		<test-containers.version>1.16.3</test-containers.version>
 		<maven-flatten-plugin.version>1.2.5</maven-flatten-plugin.version>
 		<maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
+		<mock-server.version>5.13.2</mock-server.version>
```
onobc, line 52: I am preparing to make all apps use same mock server version as there are several different versions being used currently.

functions/function/twitter-function/src/test/java/org/springframework/cloud/fn/twitter/users/TwitterUsersFunctionTests.java
```diff
@@ -184,8 +183,7 @@ public void testOne() throws IOException {
 		}
 	}
 
-	@SpringBootConfiguration
-	@EnableAutoConfiguration
+	@SpringBootApplication
```
artembilan, line 186: I think that pattern was used to avoid classpath scanning which is present on the `@SpringBootApplication`.
Not sure though if there is a hard reason to avoid such a scanning in this test suite...

functions/function/twitter-function/src/test/java/org/springframework/cloud/fn/twitter/users/TwitterUsersFunctionTests.java
artembilan: I'm not sure that syntax is correct. That's why it fails.
I believe it has to be like this: `$..code` - meaning "all codes".
See examples and their explanation here: https://github.com/json-path/JsonPath#path-examples

stream-applications-build/pom.xml
```diff
@@ -40,7 +40,7 @@
 		<nohttp-checkstyle.version>0.0.2.RELEASE</nohttp-checkstyle.version>
 		<disable.nohttp.checks>true</disable.nohttp.checks>
 		<spring-javaformat-checkstyle.version>0.0.7</spring-javaformat-checkstyle.version>
-		<spring-boot.version>2.6.7</spring-boot.version>
+		<spring-boot.version>2.7.0</spring-boot.version>
```
artembilan, line 43: What our version is this for?
Are we really planning a new minor version?
I mean it looks like there are some breaking changes pulled by this Spring Boot upgrade...

Well, it might not be a big deal for out-of-the-box apps, but if our functions are used in the end-user code they need to understand consequences.

artembilan (changes requested):
Nothing critical though.

Thanks

functions/function/twitter-function/src/test/java/org/springframework/cloud/fn/twitter/users/TwitterUsersFunctionTests.java
onobc: Hmm. It was working before so maybe they fixed a bug that was allowing that to work.

artembilan: That's my thought, too. 
Even if it worked before it doesn't mean that `..[*]` syntax is correct.

stream-applications-build/pom.xml
onobc: This is for the last release of `2021.1.x` line so the thinking was that we would get on the latest (non 3.x) Spring Boot and Spring Cloud (2021.0.3). 

Good point on the breaking property changes. Let's chat more at apps discussion today.

Issue #276, opened 2022-05-24
Enable components for customization via @CustomizationAware


onobc:
Need to decide which components to "enable" and then mark them w/ `@CustomizationAware`. 

See FileSupplierConfiguration.java for example.

Issue #275, opened 2022-05-24
Add docs for ComponentCustomizers


onobc:

onobc:
Possibly auto-generate something via the docs plugin.

onobc:
Because the component customizer is a generic interface that is shared across all of the functions, auto-generating is not something that makes sense to do at this point. 

Instead of auto-generating let's add manual documentation where it makes sense on the functions that have component customizers.

PR #281, opened 2022-05-25
Remove auto-config for `ComponentCustomizer`

artembilan:
Since we still need to modify every single function, consumer & supplier
for previously introduced `@CustomizationAware`, it does not do too much
difference for us if we add a `ComponentCustomizer` dependency to the
bean for customization and call it manually.
There is not going to be a difference for end-user to declare a
`ComponentCustomizer` bean, unless they don't need to worry about
its `static` nature anymore.

* Remove `ComponentCustomizerBeanPostProcessor` and
 respective `ComponentCustomizationAutoConfiguration`
* Use `@Nullable ComponentCustomizer<FileInboundChannelAdapterSpec>`
as dependency in the `FileSupplierConfiguration.fileMessageSource()` bean
* Rename `@CustomizationAware` to `@CustomizationDoc` since we still need
to generate docs for the customizer end-user can add for this or that function
application.

So, may benefit of this change is to avoid a `static` confusion and performance
overhead with extra `BeanPostProcessor`


functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
```diff
@@ -16,23 +16,18 @@
...
  *
  * @author Artem Bilan
  *
  * @since 1.2.1
  */
-@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
-@Retention(RetentionPolicy.RUNTIME)
-@Documented
-public @interface CustomizationAware {
+@Target(ElementType.METHOD)
```
onobc, line 30: Is the default retention sufficient for the docs plugin to do what it needs to do?

functions/common/config-common/src/main/resources/META-INF/spring.factories
```diff
@@ -1,3 +1,2 @@
 org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
-  org.springframework.cloud.fn.common.config.SpelExpressionConverterConfiguration,\
-  org.springframework.cloud.fn.common.config.ComponentCustomizationAutoConfiguration
+  org.springframework.cloud.fn.common.config.SpelExpressionConverterConfiguration
```
sobychacko, line 2: Is it no longer auto-configured? How do we provide it to individual function/app modules?

functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
```diff
@@ -16,23 +16,18 @@
...
+public @interface CustomizationDoc {
```
onobc, line 31: Do we envision ever using this annotation to signal something besides generating a doc? If so, then the name `CustomizationDoc` may be limiting.

functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
artembilan: I have no idea. 
But in my feeling this has nothing to do with `runtime` any more.

We always can fix it respectively. 😄

functions/supplier/file-supplier/src/main/java/org/springframework/cloud/fn/supplier/file/FileSupplierConfiguration.java
```diff
@@ -93,10 +95,17 @@ else if (this.fileSupplierProperties.getFilenameRegex() != null) {
 	}
 
 	@Bean
-	@CustomizationAware
-	public FileInboundChannelAdapterSpec fileMessageSource(FileListFilter<File> fileListFilter) {
-		return Files.inboundAdapter(this.fileSupplierProperties.getDirectory())
+	@CustomizationDoc
```
sobychacko, line 98: Why are we adding this customization only here? How come not in others? Or is it more like we just picked this one as an example and will do it for others later on?

functions/common/config-common/src/main/resources/META-INF/spring.factories
artembilan: Manually. See my change to the `FileSupplierConfiguration`.
Also see my commit comment for more reasoning.

functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
onobc: Also, since we are not doing any docs in this PR and we no longer need this annotation I would be in favor of removing this from the current proposal and adding what is needed when we do the docs work.

functions/supplier/file-supplier/src/main/java/org/springframework/cloud/fn/supplier/file/FileSupplierConfiguration.java
artembilan: That's correct.
If you want me to do more, then you likely need to wait more for a final fix 😄

functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
onobc: Oh no, I think this is a great simplifying improvement. The fact that we have to mark the component anyways makes total sense to update the signature to take optional customizer bean. It also clearly shows there is only 1 customizer bean that can affect the component. 

I was suggesting we can wait to add the annotation until the docs story though.

functions/supplier/file-supplier/src/main/java/org/springframework/cloud/fn/supplier/file/FileSupplierConfiguration.java
onobc: The issue for enabling more components is #276

functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
artembilan: No problem! Let's hear from others then!
@sobychacko , @corneil .

BTW, I modified JavaDocs as well to reflect reality.
So, I'm not sure what is the concern so far...

onobc (changes requested):
@artembilan I like this simplification. I put in "requested changes" for possible `CustomizationDoc` annotation removal and question on bean method "Public API". Neither of these are a must-have though.

functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
onobc: > BTW, I modified JavaDocs as well to reflect reality.

Yes, the simplification of what was needed to be doc'ed now to explain the feature is also a signal of how this has simplified things. 

> So, I'm not sure what is the concern so far...

I'm not following you on this. Can you clarify?

artembilan: I mean that annotation is now for only docs - nothing more.
I'm not sure that can figure out something else to infer the bean method info for docs generation...

But as you suggest I can remove it for the future consideration.
Not a big deal.

onobc: Another option would be finding any bean that took a nullable componentcustomizer in its bean method and use that to generate the doc for it. The downside w/ that approach is  that we would have limited context to actually doc whereas w/ an annotation we could add attributes to drive the docs etc.. 

I think we will end up w/ the annotation but we could wait to add it in as part of #275 and not have to think/decide about it in this PR.

functions/supplier/file-supplier/src/main/java/org/springframework/cloud/fn/supplier/file/FileSupplierConfiguration.java
```diff
@@ -93,10 +95,17 @@ else if (this.fileSupplierProperties.getFilenameRegex() != null) {
...
+	public FileInboundChannelAdapterSpec fileMessageSource(FileListFilter<File> fileListFilter,
```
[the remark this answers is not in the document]
artembilan: I would bite a bullet and go ahead with this breaking change.
I believe we have done similar in the past without not objections from end-users.

It is really hard to imagine that someone extends our `@Configuration` classes for their purposes...

But if you insist for consistency, then indeed such a change should go only to the next version.

We have now an auto-configuration committed which provides a already a smooth end-user experience.
And yes: that comes without breaking changes in the target configurations at all...

I know that we try to make this a extendable library, but so far we deal only with closed box components to be available in the final just via auto-configuration.

So, probably better to have a consolidated team decision without my personal preferences.

functions/common/config-common/src/main/resources/META-INF/spring.factories
sobychacko: Ah ok. That makes sense.

functions/common/config-common/src/main/java/org/springframework/cloud/fn/common/config/CustomizationDoc.java
sobychacko: That sounds like a plan to me.

functions/supplier/file-supplier/src/main/java/org/springframework/cloud/fn/supplier/file/FileSupplierConfiguration.java
sobychacko: :-) That is fine, Artem!

sobychacko (approved):

functions/supplier/file-supplier/src/main/java/org/springframework/cloud/fn/supplier/file/FileSupplierConfiguration.java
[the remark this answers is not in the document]
onobc: I vote for the approach outlined in this proposal rather than the previous auto-configuration based approach. 

I value the simplicity of this approach over the the "risk" of breaking the bean method w/ extra param. It is a very minor break if in fact someone is using the bean method in that manner.

onobc (approved):

PR #284, opened 2022-05-27
GH-276 Apply ComponentCustomizer in other modules

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/276

Some changes are straightforward as just a `ComponentCustomizer<?>`
injection and its optional usage for the target component customization.
Other changes are more drastic since it is better to use `proxyBeanMethods = false`,
so some direct bean method calls had to be reworked to the bean method argument
injections

* Fix some deprecations from Project Reactor
* Remove `UpdatingMongoDbMessageSource` since `MongoDbMessageSource`
supports now an `update` option in Spring Integration

NOTE: The `ZeroMqSupplierConfiguration` has its own customizer already
like `Consumer<ZMQ.Socket>`, so it can be revised to a more broader solution,
but in the future version.

Some modules just does not make sense to modify for more advance customization
since all their configuration is covered with plain configuration properties.
See `splitter-function` or `wavefront-consumer`


artembilan:
I believe there is no pressure from end-users, so we may safely consider this change for the next version.
Another thought: we may go baby steps and PR only for individual modules. The review and context of the changes is going to be much easier to track.

Thanks

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
```diff
@@ -41,17 +43,23 @@
...
 
 	@Bean
 	@ConditionalOnMissingBean(WebClient.class)
-	public WebClient webClient(HttpRequestFunctionProperties properties) {
-		return WebClient.builder()
-				.codecs(clientCodecConfigurer ->
-						clientCodecConfigurer.defaultCodecs().maxInMemorySize(properties.getMaximumBufferSize()))
-				.build();
+	public WebClient webClient(HttpRequestFunctionProperties properties,
+			@Nullable ComponentCustomizer<WebClient.Builder> webClientComponentCustomizer) {
```
onobc, line 53: This is a case where there is already a customizer in Boot (`org.springframework.boot.web.reactive.function.client.WebClientCustomizer`). Should we just respect that as that is what most user's would probably be familiar with?

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
```diff
@@ -24,11 +24,13 @@
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
```
onobc, line 25: **[NIT]** copyright ⬆️

functions/function/aggregator-function/src/main/java/org/springframework/cloud/fn/aggregator/AggregatorFunctionConfiguration.java
```diff
@@ -83,7 +85,8 @@ public AggregatorFactoryBean aggregator(
 			ObjectProvider<ReleaseStrategy> releaseStrategy,
 			ObjectProvider<MessageGroupProcessor> messageGroupProcessor,
 			ObjectProvider<MessageGroupStore> messageStore,
-			@Qualifier("outputChannel") MessageChannel outputChannel) {
+			@Qualifier("outputChannel") MessageChannel outputChannel,
+			@Nullable ComponentCustomizer<AggregatorFactoryBean> aggregatorCustomizer) {
```
onobc, line 89: For consistency should we use an `ObjectProvider` in this particular case since all of the other deps are using that rather than `@Nullable`?

functions/consumer/log-consumer/src/main/java/org/springframework/cloud/fn/consumer/log/LogConsumerConfiguration.java
```diff
@@ -35,16 +35,15 @@
...
  */
-@Configuration
+@Configuration(proxyBeanMethods = false)
 @EnableConfigurationProperties(LogConsumerProperties.class)
 public class LogConsumerConfiguration {
 
 	@Bean
 	IntegrationFlow logConsumerFlow(LogConsumerProperties logSinkProperties) {
 		return IntegrationFlows.from(MessageConsumer.class, (gateway) -> gateway.beanName("logConsumer"))
-				.handle((payload, headers) -> payload)
```
onobc, line 45: This must be one of the "spring integration API changes" mentioned in the PR description.

functions/consumer/rsocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/rsocket/RsocketConsumerConfiguration.java
```diff
@@ -34,17 +34,15 @@ public class RsocketConsumerConfiguration {
 	@Bean
 	public Function<Flux<Message<?>>, Mono<Void>> rsocketConsumer(RSocketRequester.Builder builder,
 																RsocketConsumerProperties rsocketConsumerProperties) {
-		final Mono<RSocketRequester> rSocketRequester =
-				rsocketConsumerProperties.getUri() != null ? builder.connectWebSocket(rsocketConsumerProperties.getUri()).cache() :
```
onobc, line 38: Ahhh yeh, this is a nice simplification of the RSocket API - they dropped the Mono. 👍🏻

functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerConfiguration.java
```diff
@@ -56,8 +59,13 @@ public IntegrationFlow ftpOutboundFlow(SftpConsumerProperties properties,
 		if (properties.getFilenameExpression() != null) {
 			handlerSpec.fileNameExpression(properties.getFilenameExpression());
 		}
+
+		if (sftpMessageHandlerSpecCustomizer != null) {
+			sftpMessageHandlerSpecCustomizer.customize(handlerSpec, "sftpMessageHandler");
```
onobc, line 64: Does "beanName" have to be an actual bean name?

functions/function/aggregator-function/src/main/java/org/springframework/cloud/fn/aggregator/AggregatorFunctionConfiguration.java
artembilan: I just didn’t know back in days that `@Nullable` will work. Or it just didn’t work in those old Spring versions… 😅

functions/supplier/ftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/ftp/FtpSupplierConfiguration.java
```diff
@@ -123,19 +132,24 @@ public Flux<Message<?>> ftpMessageFlux() {
...
+	public Supplier<Flux<Message<?>>> ftpSupplier(@Nullable Publisher<Message<Object>> ftpReadingFlow) {
 		if (this.fileConsumerProperties.getMode() == FileReadingMode.ref) {
 			return this::ftpMessageFlux;
 		}
+		else if (ftpReadingFlow != null) {
+			return () -> Flux.from(ftpReadingFlow);
+		}
 		else {
-			return () -> Flux.from(ftpReadingFlow());
+			throw new BeanInitializationException(
```
onobc, line 150: Nice addition!

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
artembilan: Yeah…  In Spring Integration I have Gradle task which updates it for me automatically for all the modified files when I compile project. So, bad habit to forgot 😅.

on the other hand: no claims that this is going to be merged in 2022. So, I’ll just wait for the final decision  😉

functions/consumer/rsocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/rsocket/RsocketConsumerConfiguration.java
artembilan: Well, it still there, but handled internally.

functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerConfiguration.java
artembilan: No. You have better ideas? 🤗

functions/supplier/ftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/ftp/FtpSupplierConfiguration.java
```diff
@@ -123,19 +132,24 @@ public Flux<Message<?>> ftpMessageFlux() {
 
 	@Bean
 	@ConditionalOnExpression("environment['file.consumer.mode'] != 'ref'")
-	public Publisher<Message<Object>> ftpReadingFlow() {
+	public Publisher<Message<Object>> ftpReadingFlow(FtpInboundFileSynchronizingMessageSource ftpMessageSource) {
```
onobc, line 135: It was not clear to me at first why we do not pass in the `FtpInboundChannelAdapterSpec ftpMessageSource` and then call `ftpMessageSource.get()`. Is it because the framework will have already called `get()` and pass us the resulting message source?

sobychacko:
@artembilan Thank you for this PR! I glanced at the changes, they look good. Once @onobc comments are addressed, I guess we can merge this one.

functions/consumer/log-consumer/src/main/java/org/springframework/cloud/fn/consumer/log/LogConsumerConfiguration.java
artembilan: Sorry, missed this one. Will add in the next commit. The `log()` now is not a terminal operator in the flow.

functions/supplier/mongodb-supplier/src/main/java/org/springframework/cloud/fn/supplier/mongo/UpdatingMongoDbMessageSource.java
```diff
@@ -1,118 +0,0 @@
...
-import org.springframework.messaging.Message;
-import org.springframework.util.Assert;
-
-/**
- * An {@link AbstractMessageSource} extension for MongoDB updates
- * based on the query result from the {@link MongoDbMessageSource} delegate.
- *
- * @author Artem Bilan
- */
-class UpdatingMongoDbMessageSource extends AbstractMessageSource<Object> {
```
onobc, line 45: Why was this deleted?

functions/supplier/ftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/ftp/FtpSupplierConfiguration.java
artembilan: Correct. That’s how I designed SI Java DSL: the `Spec` is a `FactoryBean` by itself, so no need to call it’s result manually and in some cases even dangerous. Perhaps that `get()` has to be hidden somehow, but fully different story…

when we autowire on the other side, we have to deal with the result of the `FactoryBean`. Not sure if Spring describes that somewhere, but I just share our experience what we have so far

functions/supplier/mongodb-supplier/src/main/java/org/springframework/cloud/fn/supplier/mongo/UpdatingMongoDbMessageSource.java
artembilan: I added that in the commit message

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
onobc: > no claims that this is going to be merged in 2022
Bahaha. Fair enough.

artembilan:
> I guess we can merge this one.

sorry, Soby, not today already: answering from the phone.

functions/function/aggregator-function/src/main/java/org/springframework/cloud/fn/aggregator/AggregatorFunctionConfiguration.java
onobc: Makes sense. Should we keep it consistent in this case and move to ObjectProvider to "fit in" w/ the other params?

functions/consumer/rsocket-consumer/src/main/java/org/springframework/cloud/fn/consumer/rsocket/RsocketConsumerConfiguration.java
onobc: Yep, it has to still be there but now the API  avoid the cookie cutter code that everyone did for Mono.cache->flatMap

functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerConfiguration.java
onobc: I noticed that the other specs we call `id` and then that is also what we pass into the customizer. If its simply symbolic name this is fine w/ me.

functions/supplier/ftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/ftp/FtpSupplierConfiguration.java
onobc: Thx for the explanation.

functions/supplier/mongodb-supplier/src/main/java/org/springframework/cloud/fn/supplier/mongo/UpdatingMongoDbMessageSource.java
onobc: DOH! Yes you did. My bad. 

> Remove UpdatingMongoDbMessageSource since MongoDbMessageSource
supports now an update option in Spring Integration

I did read the description (but forgot it all by the time  I got all the way to this file). I should have went back to the description before commenting.

functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerConfiguration.java
artembilan: We probably need to look at this from end-user perspective to give them an easy way to make a sense from the names we can provide them. 

I’ll give this more thoughts over night…

onobc (changes requested):
Looks good @artembilan . Thx for taking the time to move to `proxyBeanMethods = false`. I don't mid the extra effort to review it. 

My "changes requested" are minor and up for discussion anyways: 
- nullable vs object provider
- copyright
- weblient customizer

functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerConfiguration.java
artembilan: So, the name we provide to the customizer for end-user consideration must really be a real bean name.
Who knows may be they would like to have an access to that bean eventually...

Or... for consistency with Spring Boot customizer API we just can drop the `beanName` argument altogether...

onobc: If we do not have a use case yet that would require the bean name, I would be in favor of dropping it altogether (until needed). 

1. makes it consistent w/ the regular "customizer" contract (as you pointed out) 
2. makes it a bit simpler

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
onobc: Will be handled in #286

onobc (approved):
@artembilan the only outstanding item is whether or not to yank the bean name from the customizer contract. I approve the PR regardless of your choice on that. Thanks for the updates.

onobc (approved):

PR #295, opened 2022-05-30
Revise HttpRequestFunctionConfiguration configuration

### LINKED ISSUE #286: Revise HttpRequestFunctionConfiguration configuration in favor of auto-configured by Spring Boot WebClient.Builder
Currently we have a bean in the `HttpRequestFunctionConfiguration` like:
```
	@Bean
	@ConditionalOnMissingBean(WebClient.class)
	public WebClient webClient(HttpRequestFunctionProperties properties,
			ObjectProvider<WebClientCustomizer> customizerProvider) {

		WebClient.Builder builder =
				WebClient.builder()
						.codecs((clientCodecConfigurer) ->
								clientCodecConfigurer.defaultCodecs()
										.maxInMemorySize(properties.getMaximumBufferSize()));
		customizerProvider.orderedStream().forEach((customizer) -> customizer.customize(builder));
		return builder.build();
	}
```

The Spring Boot provides for us a `WebClient.Builder` auto-configuration, including those `WebClientCustomizer` s:
```
	@Bean
	@Scope("prototype")
	@ConditionalOnMissingBean
	public WebClient.Builder webClientBuilder(ObjectProvider<WebClientCustomizer> customizerProvider) {
		WebClient.Builder builder = WebClient.builder();
		customizerProvider.orderedStream().forEach((customizer) -> customizer.customize(builder));
		return builder;
	}
```

That `maxInMemorySize` property is also auto-configured by the `DefaultCodecsConfiguration`  via `CodecProperties`:
```

	/**
	 * Limit on the number of bytes that can be buffered whenever the input stream needs
	 * to be aggregated. This applies only to the auto-configured WebFlux server and
	 * WebClient instances. By default this is not set, in which case individual codec
	 * defaults apply. Most codecs are limited to 256K by default.
	 */
	private DataSize maxInMemorySize;
```

So, we can get rid of our `HttpRequestFunctionProperties.maximumBufferSize` and use already auto-configured  `WebClient.Builder` as a dependency in the `httpRequestFunction` bean definition.

sunny151091:
Fixes #286

### DIFF
--- a/applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
+++ b/applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
@@ -1,3 +1,3 @@
-configuration-properties.classes=org.springframework.cloud.fn.http.request.HttpRequestFunctionProperties
-configuration-properties.names=spring.codec.max-in-memory-size
+configuration-properties.classes=org.springframework.cloud.fn.http.request.HttpRequestFunctionProperties,\
+  org.springframework.boot.autoconfigure.codec.CodecProperties
 
--- a/applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
+++ b/applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
@@ -1,2 +1,3 @@
 configuration-properties.classes=org.springframework.cloud.fn.http.request.HttpRequestFunctionProperties
+configuration-properties.names=maximumBufferSize
 
--- a/applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
+++ b/applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
@@ -1,3 +1,3 @@
 configuration-properties.classes=org.springframework.cloud.fn.http.request.HttpRequestFunctionProperties
-configuration-properties.names=maximumBufferSize
+configuration-properties.names=spring.codec.max-in-memory-size
 
--- a/functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
+++ b/functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
@@ -46,17 +46,8 @@ import org.springframework.web.util.UriBuilderFactory;
 public class HttpRequestFunctionConfiguration {
 
 	@Bean
-	@ConditionalOnMissingBean(WebClient.class)
-	public WebClient webClient(HttpRequestFunctionProperties properties) {
-		return WebClient.builder()
-				.codecs(clientCodecConfigurer ->
-						clientCodecConfigurer.defaultCodecs().maxInMemorySize(properties.getMaximumBufferSize()))
-				.build();
-	}
-
-	@Bean
-	public HttpRequestFunction httpRequestFunction(WebClient webClient, HttpRequestFunctionProperties properties) {
-		return new HttpRequestFunction(webClient, properties);
+	public HttpRequestFunction httpRequestFunction(WebClient.Builder webClientBuilder, HttpRequestFunctionProperties properties) {
+		return new HttpRequestFunction(webClientBuilder, properties);
 	}
 
 	/**
@@ -70,8 +61,8 @@ public class HttpRequestFunctionConfiguration {
 
 		private final HttpRequestFunctionProperties properties;
 
-		public HttpRequestFunction(WebClient webClient, HttpRequestFunctionProperties properties) {
-			this.webClient = webClient;
+		public HttpRequestFunction(WebClient.Builder webClientBuilder, HttpRequestFunctionProperties properties) {
+			this.webClient = webClientBuilder.build();
 			this.properties = properties;
 		}
 
--- a/functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
+++ b/functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 the original author or authors.
+ * Copyright 2018-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,7 +22,6 @@ import java.util.function.Function;
 
 import reactor.core.publisher.Flux;
 
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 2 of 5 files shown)


applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
```diff
@@ -1,2 +1,3 @@
 configuration-properties.classes=org.springframework.cloud.fn.http.request.HttpRequestFunctionProperties
+configuration-properties.names=maximumBufferSize
```
onobc, line 2: The format for the value is the full path to the alternate config prop, which in this case is [spring.codec.max-in-memory-size](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/codec/CodecProperties.java#L28). 

You can take a look at another example in the JdbcSource app I pointed to as an example in my original reply.

onobc: There is also another "eventually-to-be-deprecated" with a similar name in this directory, can you please update it as well?

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionProperties.java
```diff
@@ -52,13 +52,6 @@ public class HttpRequestFunctionProperties {
 	 */
 	private long timeout = 30_000;
 
-	/**
-	 * Maximum buffer size in bytes allocated for input stream buffers. Defaults to 256k.
-	 * Increase, as necessary, for posting or getting large binary content.
-	 */
-
-	private int maximumBufferSize = 256 * 1024;
```
onobc, line 60: I typically would recommend that we add a unit test to cover this removal but there is currently no coverage for this particular property as it is difficult to verify anyways. Also, we can trust that the WebClient spring.codecs property has been thoroughly tested.

onobc (changes requested):
Nice work @sunny151091 !

There are few items to address but all small things. 

One thing  I would recommend (I do this myself anytime I change an app) is to build the specific rabbit/kafka version of the app and run it locally. It's a great verification (although we do cover that case during out CI builds) and also a good leaning tool to see how the apps are actually constructed. 


The process is outlined in the readme. Just to summarize, the content in each `applications/<source|sink|processor>/<appname>` is a template that brings in some function dependencies (and other deps) and we run the app generator plugin on that which in turn creates 2 apps under `applications/<source|sink|processor>/<appname>/apps` - take a look in there a see what you see. You can then drop into the rabbit/kafka apps and build/run those and see the final app actually working like it does when its downloaded from maven/docker by end users.

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
```diff
@@ -70,8 +61,8 @@ public static class HttpRequestFunction implements Function<Flux<Message<?>>, Fl
 
 		private final HttpRequestFunctionProperties properties;
 
-		public HttpRequestFunction(WebClient webClient, HttpRequestFunctionProperties properties) {
-			this.webClient = webClient;
+		public HttpRequestFunction(WebClient.Builder webClientBuilder, HttpRequestFunctionProperties properties) {
+			this.webClient = webClientBuilder.build();
```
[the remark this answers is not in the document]
artembilan: I think there was a reason in Spring Boot to not expose a `WebClient` as a bean. Therefore having a `WebClient.Builder` injected is OK.
And then we use it internally to create a respective instance as it is intended.

This is already some breaking change. Therefore I didn't mark the issue to be included in the current version.
For the future version it is fully OK to drop our property and remove `WebClient` bean.

artembilan (changes requested):
This is great even if we have a couple concerns.

Thanks

artembilan (approved):

artembilan (approved):
I don't have any more concerns unless we would like to revise that new property to the whole `org.springframework.boot.autoconfigure.codec.CodecProperties` class to expose.

Let's see what @onobc thinks on the matter!

functions/function/http-request-function/src/main/java/org/springframework/cloud/fn/http/request/HttpRequestFunctionConfiguration.java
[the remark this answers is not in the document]
onobc: > I think there was a reason in Spring Boot to not expose a WebClient as a bean. Therefore having a WebClient.Builder injected is OK.

Agreed @artembilan . I meant just passing in the WebClient to the `HttpRequestFunction.init` as it was being done before (like [this](https://github.com/spring-cloud/stream-applications/pull/295/files#diff-f40508500c04ef15efaf5299bc78fcdda9485a34386a76c0af0b75aa903151dfR52)).

applications/processor/http-request-processor/src/main/resources/META-INF/dataflow-configuration-metadata.properties
onobc: @artembilan If all of the properties apply to the `WebClient` then that would be a great thing to do. 

@sunny151091 - this adjustment would result in all of the properties in `CodecProperties ` to be listed in the app docs (vs. only the single one that I suggested to add initially).

onobc (changes requested):
Thanks for the updates @sunny151091 . Looking good. 

These are the only 2 items remaining: 
1. I think we should update the metadata files to use the `CodecProperties` (vs. the single property).
2. I am concerned that I do not see any updated apps docs in the PR (see my comment about this:
> When you build in the project it auto-generates docs based on these files. This change should have resulted in a change to the generated docs. I believe the doc generation will ignore any props it can not resolve which explains why there was no update. See what happens after you update it as outlined above and re-run the build.

onobc:
>Hi @onobc, have added CodecProperties. But i don't know which doc will get auto-generated after building the application.
After making the changes i re-build the http-request-function and the build was successful. However no docs were auto-generated.
Could you help me out here?

Hi @sunny151091 ,

Just run the build from the root directory as described at the top of [Build](https://github.com/spring-cloud/stream-applications/blob/main/README.adoc#build) section in the README. This should generate the docs and also execute the related app tests.

Thanks

artembilan (approved):
Hi @sunny151091 !

I'm not @onobc , but I want to say you "Thank you!" for such a great contribution!

I just want to leave this comment about my personal contribution cycle preferences.

So, you said:

> Have committed the auto generated app docs. 

However we definitely see a new commit from you therefore not need to duplicate a message on the matter. 
Unless we forget about your PR for a long time 😉 

Also: if you agree with our review, no reason to say that: just push the fix which really confirm an agreement!
Otherwise it turns to a spam in my mailbox but I have to deal with hundreds of PRs or GH issue sometime...

In addition, just for future considerations. We try to follow these Git commit rules: https://cbea.ms/git-commit/.
Of course we will fix the commit message on merge, but it would save yours and ours time during PR preparing and review, plus merging.

Thanks for understanding!

Re. your PR: all good, but we cannot merge it yet because we haven't switched `main` branch to new major or minor version. We cannot apply this change to the point version since it comes as a breaking change.

onobc (approved):

sunny151091:
Thanks a lot @onobc and @artembilan. 

This was my first ever contribution to open source. I will take a note of all the feedback you guys shared and adhere to the contribution guidelines in my future contributions.

onobc:
@sunny151091 
We will be creating a branch for the next release (`2022.0.0`) shortly and will merge this in at that point. 

Thanks again for the contribution.

PR #325, opened 2022-09-21
Disable DB init for JDBC Source and Sink

artembilan:
The `IntegrationAutoConfiguration` provides mechanism to initialize DB with some out-of-the-box scripts. The default behavior is `EMBEDDED` even if we don't use any components for those DB objects, e.g. no `JdbcMessageStore` bean.

* Add `spring.integration.jdbc.initialize-schema=NEVER` to the `jdbc-consumer` and `jdbc-supplier` to not attempt to initialize DB
* Fix Checkstyle violations in other modules

### DIFF
--- a/functions/supplier/jdbc-supplier/src/test/java/org/springframework/cloud/fn/supplier/jdbc/DefaultJdbcSupplierTests.java
+++ b/functions/supplier/jdbc-supplier/src/test/java/org/springframework/cloud/fn/supplier/jdbc/DefaultJdbcSupplierTests.java
@@ -82,6 +82,11 @@ public class DefaultJdbcSupplierTests {
 		stepVerifier.verify();
 	}
 
+	/*
+	The test to verify that DB is not initialized with Spring Integration DDL
+	(spring.integration.jdbc.initialize-schema=NEVER) what happens by default via IntegrationAutoConfiguration.IntegrationJdbcConfiguration.
+	This is not a functionality of this JDBC Supplier.
+	 */
 	@Test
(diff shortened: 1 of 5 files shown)


functions/supplier/jdbc-supplier/src/test/java/org/springframework/cloud/fn/supplier/jdbc/DefaultJdbcSupplierTests.java
```diff
@@ -76,6 +82,13 @@ void testExtraction() {
 		stepVerifier.verify();
 	}
 
+	@Test
+	void verifyNoIntMessageGroupTable() {
+		assertThatExceptionOfType(BadSqlGrammarException.class)
```
onobc, line 87: This is indirectly verifying the fact that the schema did not initialize. Might be helpful to add a 1-liner comment or include this fact in the name of the test.

onobc (changes requested):
@artembilan looks good. I left 1 nit comment/request - once addressed (or not) we can merge. Thanks.

functions/supplier/jdbc-supplier/src/test/java/org/springframework/cloud/fn/supplier/jdbc/DefaultJdbcSupplierTests.java
onobc: awesome details - thanks @artembilan

onobc (approved):

PR #338, opened 2022-10-04
Remove extension settings.xml

onobc:
A maven extension to allow project level settings.xml was previously added to fix an issue on Jenkins CI (https://github.com/spring-cloud/stream-applications/commit/9fbec999552fa1348460c806da65b6b74337e93f).

However, this is now causing issues when importing project into IDEA. Also, main will not be built on Jenkins so we are removing the extension.

Also remove the old custom `.settings.xml` in the root of the project and put the repos inline in `stream-applications-build`.


dmfrey:
Builds off of `main` seem to still be looking for a `spring` maven profile, which, I'm assuming, was only in the local maven settings.xml or .settings.xml files

onobc:
@dmfrey yes, this profile is still available on Jenkins in the default global profile. However, since we are going to Github actions we should be able to remove the spring profile from the default args in ./maven/maven.config.

dmfrey:
agreed, just highlighting it was left over. I commented it out locally and that message went away. There also still seems to be some leftover references to the geode code in there which is no longer in the branch.

onobc:
Thx @dmfrey , both of these are in the process of  being removed.

Issue #348, opened 2022-10-11
Update SftpConsumer properties to support Resource for knownHosts


onobc:
I'm not sure why that is an expression at all, but I believe that location to such a `known-hosts` file could be in the classpath, not just an external file system.
Can we revise this into something like: 
```
@Value("#{${sftp.consumer.factory.knownHostsExpression}}") Resource knownHosts
``` 
?

_Originally posted by @artembilan in https://github.com/spring-cloud/stream-applications/pull/347#discussion_r991446554_

PR #349, opened 2022-10-11
GH-348: Resolve knownHostsExpression to Resource

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/348

Currently, SFTP modules use a `FileSystemResource` based on a resolved from knownHostsExpression.
This is not always true since the resource could be anywhere, not only in the target file system.
The typical use-use is classpath or Secretes on the target platform

* Use `DefaultResourceLoader` to resolve a location from the expression to any possible `Resource` impl

### DIFF
--- a/functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerSessionFactoryConfiguration.java
+++ b/functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerSessionFactoryConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 the original author or authors.
+ * Copyright 2015-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,14 +18,14 @@ package org.springframework.cloud.fn.consumer.sftp;
 
 import org.apache.sshd.sftp.client.SftpClient;
 
-import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
-import org.springframework.core.io.FileSystemResource;
-import org.springframework.integration.context.IntegrationContextUtils;
+import org.springframework.core.io.Resource;
 import org.springframework.integration.file.remote.session.CachingSessionFactory;
 import org.springframework.integration.file.remote.session.SessionFactory;
 import org.springframework.integration.sftp.session.DefaultSftpSessionFactory;
+import org.springframework.lang.Nullable;
 
 /**
  * Session factory configuration.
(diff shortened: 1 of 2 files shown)


functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerSessionFactoryConfiguration.java
```diff
@@ -18,27 +18,31 @@
...
+ * @author Artem Bilan
  */
 public class SftpConsumerSessionFactoryConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public SessionFactory<SftpClient.DirEntry> sftpSessionFactory(SftpConsumerProperties properties, BeanFactory beanFactory) {
+	public SessionFactory<SftpClient.DirEntry> sftpSessionFactory(
+			SftpConsumerProperties properties,
+			@Value("#{${sftp.consumer.factory.known-hosts-expression:null}}") @Nullable Resource knownHosts) {
```
onobc, line 44: This confused me for a second, as I was expecting the expression to be on the properties. However, it is in fact the path `sftp.consumer.factory.known-hosts-expression` to the field on the properties. I see your reason for doing this - let Spring inject it and we not have to write the following code :
```java
if (factory.getKnownHostsExpression() != null) {
  String location = factory.getKnownHostsExpression()
	  .getValue(IntegrationContextUtils.getEvaluationContext(beanFactory), String.class);
  sftpSessionFactory.setKnownHostsResource(resourceLoader.getResource(location));
}
```

functions/supplier/sftp-supplier/src/main/java/org/springframework/cloud/fn/supplier/sftp/SftpSupplierFactoryConfiguration.java
```diff
@@ -84,9 +85,10 @@ static SessionFactory<SftpClient.DirEntry> buildFactory(BeanFactory beanFactory,
 		sftpSessionFactory.setPrivateKeyPassphrase(factory.getPassPhrase());
 		sftpSessionFactory.setAllowUnknownKeys(factory.isAllowUnknownKeys());
 		if (factory.getKnownHostsExpression() != null) {
-			String path = factory.getKnownHostsExpression()
+			String knownHostsLocation = factory.getKnownHostsExpression()
 					.getValue(IntegrationContextUtils.getEvaluationContext(beanFactory), String.class);
-			sftpSessionFactory.setKnownHostsResource(new FileSystemResource(path));
+			Resource knownHostsResource = new DefaultResourceLoader().getResource(knownHostsLocation);
```
onobc, line 90: Any reason not to use the SpEL method param to get the resource like you are doing above for `SftpConsumerSessionFactoryConfiguration`?  I am ok w/ either approach but for consistency, can they be the same?

Also, if we stick w/ the programmatic route, wdyt about using the container's `ResourceLoader` (either via method param injection or via ResourceLoaderAware)?

functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerSessionFactoryConfiguration.java
artembilan: I think `@Resource` has its own limitations in Spring and technically it is more an equivalent to the `@Autowired`, not `@Value`.
And definitely it has nothing to do with the `org.springframework.core.io.Resource` 😄

functions/consumer/sftp-consumer/src/main/java/org/springframework/cloud/fn/consumer/sftp/SftpConsumerSessionFactoryConfiguration.java
```diff
@@ -52,7 +52,12 @@ public SessionFactory<SftpClient.DirEntry> sftpSessionFactory(
 		sftpSessionFactory.setPrivateKey(factory.getPrivateKey());
 		sftpSessionFactory.setPrivateKeyPassphrase(factory.getPassPhrase());
 		sftpSessionFactory.setAllowUnknownKeys(factory.isAllowUnknownKeys());
-		sftpSessionFactory.setKnownHostsResource(knownHosts);
+		if (factory.getKnownHostsExpression() != null) {
+			String knownHostsLocation = factory.getKnownHostsExpression()
+					.getValue(IntegrationContextUtils.getEvaluationContext(applicationContext), String.class);
+			Resource knownHostsResource = applicationContext.getResource(knownHostsLocation);
```
onobc, line 58: Ahh yes, this is even better, I forget that AC is actually an RL!

onobc (approved):
Looks good @artembilan . Thx for the fix!

Issue #369, opened 2022-10-18
HeaderEnricherProcessor/Function fails


onobc:
The HeaderEnricher functions/apps fails on "unable to resolve '@value'" as the bean factory is null for some reason. 

- https://github.com/spring-cloud/stream-applications/commit/b8571cba4f0b411770fedd710afac9fb9abb8bd1
- https://github.com/spring-cloud/stream-applications/commit/d351163cd2cc3ee16a3ca1e3608d6ee676db262e

PR #376, opened 2022-10-19
GH-369: Fix HeaderEnricher for no proxyBeanMethod

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/369

Spring Boot `@AutoConfiguration` comes now with a `proxyBeanMethods = false`, so we cannot call bean methods within the same configuration class

* Rework `HeaderEnricherFunctionConfiguration` for bean methods autowiring
* Create an `ExpressionEvaluatingHeaderValueMessageProcessor` instances directly in the `headerEnricher` bean definition and propagate an injected `BeanFactory`

### DIFF
--- a/functions/function/header-enricher-function/src/test/java/org/springframework/cloud/fn/header/enricher/HeaderEnricherFunctionApplicationTests.java
+++ b/functions/function/header-enricher-function/src/test/java/org/springframework/cloud/fn/header/enricher/HeaderEnricherFunctionApplicationTests.java
@@ -18,7 +18,6 @@ package org.springframework.cloud.fn.header.enricher;
 
 import java.util.function.Function;
 
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


functions/function/header-enricher-function/src/test/java/org/springframework/cloud/fn/header/enricher/HeaderEnricherFunctionApplicationTests.java
```diff
@@ -42,7 +41,6 @@
 		"header.enricher.headers=foo='bar' \\n baz='fiz' \\n buz=payload \\n jaz=@value",
 		"header.enricher.overwrite = true" })
 @DirtiesContext
-@Disabled
```
onobc, line 45: Can you also re-enable 
`org.springframework.cloud.stream.app.processor.header.enricher.HeaderEnricherProcessorTests`? I missed this one in the issue description. I can also do it after this is merged if you are busy. Thanks for the quick fix @artembilan . I'm glad it was straightforward.

onobc (changes requested):

onobc (approved):

PR #380, opened 2022-10-21
move xmpp testcontainer to function-test-support

dmfrey:
This will move the XMPP TestContainer setup and configuration to the `function-test-support` library.

There is now an abstract class that will set up, configure, and starts the XMPP test container. It also exposes values for the registered test users defined by the XMPP server configuration. These can be used to configure the server connection and a client connection (if needed) and can be added to a Spring `DynamicPropertySource` configuration through an annotation.

Since the configuration is now located in a Jar file, the typical `ClassPathResource(...).getFile()` will *NOT* be able to find the file. This is only an issue when executing from the command line. The IDE does not have this problem.

To solve this problem, the Maven plugin `maven-dependency-plugin` is used to copy the required resources out of the jar during the `test-compile` phase into the `${project.build.testOutputDirectory}`. Once complete, the library can look up the file easily as a `ClassPathResource`.


functions/consumer/xmpp-consumer/src/test/java/org/springframework/cloud/fn/consumer/xmpp/XmppConsumerConfigurationTests.java
```diff
@@ -163,13 +144,29 @@ private void assertStanza(Stanza stanza) {
...
 	}
 
 	private void assertFrom(Stanza stanza) {
 
-		assertThat(stanza.getFrom().asBareJid().asUnescapedString()).isEqualTo(FROM + "@" + SERVICE_NAME);
+		assertThat(stanza.getFrom().asBareJid().asUnescapedString()).isEqualTo(JOHN_USER + "@" + SERVICE_NAME);
+
+	}
+
+	static class Initializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
```
artembilan, line 157: See `@DynamicPropertySource` instead.

artembilan (changes requested):

functions/common/function-test-support/src/main/java/org/springframework/cloud/fn/test/support/xmpp/XmppContainerSupport.java
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	public static final String SERVICE_NAME = "localhost";
+
+	static final DockerComposeContainer XMPP_CONTAINER;
+
+	static {
+
+		try {
+
+			File dockerComposeFile = new ClassPathResource("/docker-compose/xmpp/docker-compose.yml").getFile();
```
onobc, line 65: In relation to the fact of having to extract the content from the jar in order to reference it as a file...

I would really like to get rid of the maven plugin requirement. 

#### Option 1 - copy resource via in/out streams

To access the resource in all cases we have to get the input stream from the resource and then copy/transfer to the output stream. It looks something like this:

```java
File dockerComposeFile = File.createTempFile("my-docker-compose", ".yml");
ClassPathResource resource = new ClassPathResource("docker-compose/xmpp/docker-compose.yml");
try (InputStream inputStream = resource.getInputStream()) {
	try (FileOutputStream outputStream = new FileOutputStream(dockerComposeFile, false)) {
	  inputStream.transferTo(outputStream);
	  // OR FileCopyUtils.copy(inputStream, outputStream);
	}
}
System.out.println("***** " + dockerComposeFile.getAbsolutePath());
```
That is easy enough (works in both IDE and CLI), but the next issue is all of the conf that the docker-compose points to w/ its volume mount - it too has to have the same treatment. It can be done, but that is getting a bit complicated. 

#### Option 2 - create custom container / use built-in api 

There is a built-in API in `GenericContainer` that handles this for us `withClasspathResourceMapping`. We can create a custom XmppContainer, delete all use of docker-compose, solve the dynamic port issue, and the reosurce loading issue all in one shot. This got me really interested and I put together a WIP branch to illustrate. Please have a look and lmk what you think. https://github.com/spring-cloud/stream-applications/pull/382

functions/common/function-test-support/src/main/resources/docker-compose/xmpp/docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 version: "3.7"
 services:
-  Openfire:
+  xmpp:
     image: fishbowler/openfire:latest
```
onobc, line 4: It may be less surprised down the road if we lock onto a specific version (`v4.7.0` seems to be a good candidate).

onobc (changes requested):
@dmfrey I like the direction this is going - nice work. I just realized my summary comment is intertwined in my comment on [XmppContainerSupport.java](https://github.com/spring-cloud/stream-applications/pull/380#discussion_r1002756997).

functions/common/function-test-support/src/main/java/org/springframework/cloud/fn/test/support/xmpp/XmppContainerSupport.java
```diff
@@ -0,0 +1,79 @@
...
+package org.springframework.cloud.fn.test.support.xmpp;
+
+import java.io.File;
+import java.io.IOException;
+
+import org.testcontainers.containers.DockerComposeContainer;
+
+import org.springframework.core.io.ClassPathResource;
+
+public abstract class XmppContainerSupport {
```
onobc, line 26: I am sure you noticed the barrage of logging from the testcontainer now when running tests. To get rid of it you can create a `/stream-applications/functions/common/function-test-support/src/main/resources/logback-test.xml` that looks like this: 

```xml
<configuration>
    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
        <encoder>
            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger - %msg%n</pattern>
        </encoder>
    </appender>
    <root level="WARN">
        <appender-ref ref="STDOUT"/>
    </root>
    <logger name="org.testcontainers" level="ERROR"/>
    <logger name="com.github.dockerjava" level="ERROR"/>
</configuration>
```

And voila, its gone.

onobc (approved):

onobc:
Closed via https://github.com/spring-cloud/stream-applications/commit/9a6f05def7f69046cd8c462db480464960559163

Issue #350, opened 2022-10-12
Fix Cassandra Consumer for latest Spring Integration 6.0


artembilan:
For more info: https://github.com/spring-projects/spring-integration/pull/3913.

Pay attention to the `CassandraContainerTest` and `IntegrationTestConfig` to setup integration testing environment.

PR #381, opened 2022-10-21
GH-350: Upgrade SI Cassandra; re-enable tests

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/350

* Use Testcontainers for Cassandra
* Adjust components code style a bit
* Convert `Book` testing entity to the `record`
* Use `@DynamicPropertySource` to expose Docker ENV to Spring Boot configuration properties

### DIFF
--- a/functions/consumer/cassandra-consumer/src/main/java/org/springframework/cloud/fn/consumer/cassandra/CassandraConsumerConfiguration.java
+++ b/functions/consumer/cassandra-consumer/src/main/java/org/springframework/cloud/fn/consumer/cassandra/CassandraConsumerConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


functions/consumer/cassandra-consumer/src/main/java/org/springframework/cloud/fn/consumer/cassandra/CassandraConsumerConfiguration.java
```diff
@@ -82,19 +88,19 @@ public IntegrationFlow cassandraConsumerFlow(MessageHandler cassandraSinkMessage
...
+		CassandraMessageHandler cassandraMessageHandler = new CassandraMessageHandler(cassandraOperations, queryType);
 		cassandraMessageHandler.setProducesReply(true);
-		cassandraMessageHandler.setAsync(true);
-		if (this.cassandraSinkProperties.getConsistencyLevel() != null
-				|| this.cassandraSinkProperties.getTtl() > 0) {
+		int ttl = this.cassandraSinkProperties.getTtl();
+		ConsistencyLevel consistencyLevel = this.cassandraSinkProperties.getConsistencyLevel();
+		if (consistencyLevel != null || ttl > 0) {
 
 			WriteOptions.WriteOptionsBuilder writeOptionsBuilder = WriteOptions.builder();
```
onobc, line 101: This is a great candidate for the enhanced switch statement - it reduces it down to this:
```java
	WriteOptions.WriteOptionsBuilder writeOptionsBuilder = switch (queryType) {
		case INSERT -> InsertOptions.builder();
		case UPDATE -> UpdateOptions.builder();
		default -> WriteOptions.builder();
	};
```

functions/consumer/cassandra-consumer/src/main/java/org/springframework/cloud/fn/consumer/cassandra/CassandraConsumerConfiguration.java
```diff
@@ -164,8 +167,7 @@ protected List<List<Object>> transformPayload(Object payload) {
 						List<Object> row = new ArrayList<>(this.columns.size());
 						for (String column : this.columns) {
 							Object value = entity.get(column);
-							if (value instanceof String) {
-								String string = (String) value;
+							if (value instanceof String string) {
```
onobc, line 170: Nice - I like the enhanced `instanceof`

onobc (approved):
Nice work @artembilan ! Excellent use of recent features/enhancements. 

I ❤️ seeing the PRs where we switch from POJO to `record` - always a huge reduction of boilerplate. 

The swap of embedded for testcontainers simplifies things greatly as well. 

I had 1 minor sugggestion/improvement. If you agree w/ that lml and I can update that during merge.

functions/consumer/cassandra-consumer/src/main/java/org/springframework/cloud/fn/consumer/cassandra/CassandraConsumerConfiguration.java
artembilan: Yeah, thought about that. But looks like I have mixed it with a pattern matching expression which is still preview: https://docs.oracle.com/en/java/javase/17/language/pattern-matching-switch-expressions-and-statements.html.
Looks like this expression one is standard since Java 14. So, feel free to fix it as you wish 😅

onobc: Thanks for update @artembilan . I was planning on getting to this some time today.

PR #382, opened 2022-10-23
[WIP - ILLUSTRATION_ONLY] Switch to custom test container

onobc:
@dmfrey


functions/common/function-test-support/pom.xml
```diff
@@ -47,8 +47,12 @@
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers</artifactId>
-            <version>${test-containers.version}</version>
+            <scope>compile</scope>
```
artembilan, line 50: But you said yourself in other place that this one is a default scope…

functions/common/function-test-support/src/main/java/org/springframework/cloud/fn/test/support/xmpp/XmppContainer.java
```diff
@@ -0,0 +1,43 @@
...
+	public XmppContainer() {
+		this(DEFAULT_IMAGE_NAME);
+	}
+
+	public XmppContainer(DockerImageName dockerImageName) {
+		super(dockerImageName);
+		dockerImageName.assertCompatibleWith(DEFAULT_IMAGE_NAME);
+		withExposedPorts(XMPP_INTERNAL_PORT);
+		withClasspathResourceMapping("xmpp/conf", "/var/lib/openfire/conf", BindMode.READ_ONLY);
+		withCommand("-demoboot");
```
artembilan, line 39: Doesn’t look like any  of these give us a justification of a custom impl. Why all of these cannot go directly to that interface you have introduced to start a generic container in tests?

functions/consumer/xmpp-consumer/src/test/java/org/springframework/cloud/fn/consumer/xmpp/XmppConsumerConfigurationTests.java
```diff
@@ -55,30 +54,19 @@
...
+public class XmppConsumerConfigurationTests implements XmppTestContainerSupport {
+
+	@DynamicPropertySource
+	static void registerConfigurationProperties(DynamicPropertyRegistry registry) {
+		registry.add("xmpp.factory.user", () -> JOHN_USER);
+		registry.add("xmpp.factory.password", () -> USER_PW);
+		registry.add("xmpp.factory.host", () -> XmppTestContainerSupport.getXmppHost());
+		registry.add("xmpp.factory.port", () -> XmppTestContainerSupport.getXmppMappedPort());
+		registry.add("xmpp.factory.service-name", () -> SERVICE_NAME);
+		registry.add("xmpp.factory.security-mode", () -> "disabled");
```
artembilan, line 68: Looks like some of these are really not dynamic. What is wrong with specifying via removed `properties`?

functions/common/function-test-support/pom.xml
onobc: I did not update the original changes - it was copied over.

functions/consumer/xmpp-consumer/src/test/java/org/springframework/cloud/fn/consumer/xmpp/XmppConsumerConfigurationTests.java
onobc: Agreed, I had a similar question in original PR - this was just copy over from the PR.

functions/common/function-test-support/src/main/java/org/springframework/cloud/fn/test/support/xmpp/XmppContainer.java
onobc: Yeh, good point. As long as we can access these methods, then that would simplify things even more.

onobc:
Superceded by https://github.com/spring-cloud/stream-applications/commit/9a6f05def7f69046cd8c462db480464960559163

Issue #206, opened 2021-12-21
DLQ enabling is not working with 'http-request'


CEDDM:
I'm using the `http-request 3.1.1` app in `SCDF 2.9.1` and I want the Kafka message to be stored in a DLQ in case the HTTP service invoked by `http-request` is not responding.
The idea is to resubmit the message when the HTTP service is back up.

I'm using this configuration :
```
--spring.cloud.stream.kafka.default.consumer.enableDlq=true
--spring.cloud.stream.kafka.default.consumer.dlqName=errorHttpRequest
```

I have this error :
```
c.f.c.c.BeanFactoryAwareFunctionRegistry : Failed to invoke function 'httpRequestFunction'
2021-12-21T11:48:40.076935527+01:00 stdout F 
2021-12-21T11:48:40.076942067+01:00 stdout F org.springframework.web.reactive.function.client.WebClientRequestException: finishConnect(..) failed: Connection refused: /<IP>:8085; nested exception is io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: /<IP>:8085
```
Which is normal but nothing in the DLQ ...

Note that this was working fine with the previous app `httpclient`

artembilan:
Can you show, please, the full stack trace?
To let us to determine the point of failure, so we probably would know more how to help you to workaround the issue.
In fact most of the functions in the current stream application generation are reactive, so not having it routed to DLQ looks expected since nothing is thrown directly to the message listener container for it error handling process.
That's my logical guess from the reactive perspective.
But a full stack trace would give us more clues what is there.

Thanks

CEDDM:
Thank you for the explanation.
Here is the full stack trace :
```
2021-12-21T15:48:08.886695136+01:00 stdout F 2021-12-21 14:48:08.882 ERROR [http-request-processor,91548655281a3f62,59606d272eb2fb73] 1 --- [or-http-epoll-2] c.f.c.c.BeanFactoryAwareFunctionRegistry : Failed to invoke function 'httpRequestFunction'
2021-12-21T15:48:08.886709206+01:00 stdout F 
2021-12-21T15:48:08.886715458+01:00 stdout F org.springframework.web.reactive.function.client.WebClientRequestException: finishConnect(..) failed: Connection refused: /<IP>:8085; nested exception is io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: /<IP>:8085
[...]

PR #395, opened 2022-10-26
Reproducing #206 with simple filter example.

corneil:


corneil:
Included with other merge.

PR #413, opened 2022-11-10
initial commit

dmfrey:
implement `xmpp-source` application which utilizes the `xmpp-supplier` to receive messages from an XMPP Server

### DIFF
--- a/applications/source/pom.xml
+++ b/applications/source/pom.xml
@@ -10,26 +10,27 @@
     <packaging>pom</packaging>
 
     <modules>
+        <module>cdc-debezium-source</module>
         <module>file-source</module>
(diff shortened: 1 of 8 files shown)


applications/source/pom.xml
```diff
@@ -10,26 +10,27 @@
     <packaging>pom</packaging>
 
     <modules>
+        <module>cdc-debezium-source</module>
```
onobc, line 13: I am guessing the only change is adding the new app? I appreciate the alpha-order, but a separate commit would make it easier to see what is formatting vs actual changes.

onobc (approved):
Nice work @dmfrey ! Thanks for another awesome contribution.

PR #434, opened 2023-02-22
Remove spring-doc-resources-0.2.5.zip.

corneil:
Fix repos for snapshots and milestones.


.settings.xml
```diff
@@ -25,15 +25,15 @@
                 <repository>
                     <id>spring-snapshots</id>
                     <name>Spring Snapshots</name>
-                    <url>https://repo.spring.io/snapshot</url>
+                    <url>https://repo.spring.io/libs-snapshot</url>
```
artembilan, line 28: Why is that?
The recommendation is really to use just `snapshot` where we have only snapshots.
Mostly artifacts must be resolved from Maven Central.
The artifact under the question is here: https://mvnrepository.com/artifact/io.spring.docresources/spring-doc-resources/0.2.5
Right, it is not in a Central by itself, but in the https://repo.spring.io/release/io/spring/docresources/spring-doc-resources/0.2.5/

So, we probably need to add that https://repo.spring.io/release entry instead of this `libs` mirror.

artembilan (changes requested):

.settings.xml
corneil: libs-snapshot includes spring milestones and releases.

artembilan: Correct, but I believe it requires an authentication.
We should do like this instead:
```
        maven { url 'https://repo.spring.io/release' }
        maven { url 'https://repo.spring.io/milestone' }
        maven { url 'https://repo.spring.io/snapshot' }
```

onobc: @corneil I thought we only needed to use the "uber" virtual repos due to JFrog plugin which we only use in SCDF.

PR #436, opened 2023-02-23
Clarify sftp-supplier docs

onobc:
* Highlight the `ref` mode invalid when streaming
* Use properties rather than shell options in docs

See https://github.com/spring-cloud/spring-cloud-dataflow/issues/5225


artembilan (approved):
The PR CI build is still failing, but I'm merging this one anyway.
Thank you!

PR #439, opened 2023-03-14
Remove invalid config props references

### LINKED ISSUE #435: Invalid config props in apps
Some apps `dataflow-configuration-metadata.properties` have references to config props that no longer exist in SB3.

One example is [ElasticSearch Sink](https://github.com/spring-cloud/stream-applications/blob/main/applications/sink/elasticsearch-sink/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties#L2) where `ElasticsearchRestClientProperties` no longer exists. It needs to be removed and maybe replaced w/ other ES props based on which client is currently being used by the apps.

Need to find out if there are other apps that have the same issue.

onobc:
* `sink/elasticsearch-sink`: SB3 now uses Elastic Java client instead of Elastic High-Level REST client, therefore replace `ElasticsearchRestClientProperties` w/ `ElasticsearchProperties.

* `processor/aggregator-processor`: SB3 has removed Embedded Mongo support, therefore remove the corresponding EmbeddedMongoProperties.

Fixes #435


artembilan (approved):
I'm OK with this, but what is that `IllegalStateException` in the GH action for PR?

onobc:
> I'm OK with this, but what is that `IllegalStateException` in the GH action for PR?

I think its unrelated and has to do w/ the change detection in the CI plan. @corneil is going to take a look when he gets online tomorrow.

PR #443, opened 2023-04-05
Added header-filter function and processor.

corneil:


corneil:
Merging because ci-pr needs fixing

PR #459, opened 2023-05-24
Factor out the Debezium Engine Builder auto-configuration

### LINKED ISSUE #458: The Debezium auto-configuration to expose the DebeziumEngine.Builder
Refactor the `DebeziumEngineAutoConfiguration` so that  it auto-creates a DebeziumEngine builder (e.g. `DebeziumEngine.Builder<ChangeEvent<byte[], byte[]>>`) instance instead of `DebeziumEngine<ChangeEvent<byte[], byte[]>>`. The creation of the later should be part of the  `DebeziumReactiveConsumerConfiguration` (non-auto) configuration.

This change would make the Debeizum supplier auto-configuration far more versatile and reusable across other Spring projects such as the Spring Integration [Debezium Input Channel Adapter](https://github.com/spring-projects/spring-integration/tree/main/spring-integration-debezium).

tzolov:
- Rename DebeziumEngineAutoConfiguration into DebeziumEngineBuilderAutoConfiguration.
 - Make the DebeziumEngineBuilderAutoConfiguration auto-create a DebeizumEngine.Builder instance.
 - Move the DebeizumEngine bean creation to the DebeziumReactiveConsumerConfiguration.
 - Make DebeziumEngineBuilderAutoConfiguration conditional on Debeizum connecgtor on the class path and debezium.properties.connector.class property set.
 - Minor improvment on the serialization format class resolution.
 - Add support for debezium header serialization format, required for the change event headers to work.

 Resolves #458

### DIFF
--- a/functions/supplier/debezium-supplier/src/test/java/org/springframework/cloud/fn/supplier/debezium/custom/DebeziumEngineBuilderAutoConfigurationTest.java
+++ b/functions/supplier/debezium-supplier/src/test/java/org/springframework/cloud/fn/supplier/debezium/custom/DebeziumEngineBuilderAutoConfigurationTest.java
@@ -25,7 +25,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
 
 import io.debezium.engine.ChangeEvent;
-import io.debezium.engine.DebeziumEngine;
+import io.debezium.engine.DebeziumEngine.Builder;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.junit.jupiter.api.Tag;
@@ -57,8 +57,8 @@ import static org.awaitility.Awaitility.await;
  */
 @Tag("integration")
 @Testcontainers
-public class DebeziumEngineAutoConfigurationTest {
-	private static final Log logger = LogFactory.getLog(DebeziumEngineAutoConfigurationTest.class);
+public class DebeziumEngineBuilderAutoConfigurationTest {
+	private static final Log logger = LogFactory.getLog(DebeziumEngineBuilderAutoConfigurationTest.class);
 
 	private static final String DATABASE_NAME = "inventory";
 	public static final String IMAGE_TAG = "2.2.0.Final";
@@ -144,8 +144,9 @@ public class DebeziumEngineAutoConfigurationTest {
 	public static class DebeziumCustomConsumerApplication {
 
 		@Bean
-		public EmbeddedEngineExecutorService embeddedEngine(DebeziumEngine<?> debeziumEngine) {
-			return new EmbeddedEngineExecutorService(debeziumEngine);
+		public EmbeddedEngineExecutorService embeddedEngine(Consumer<ChangeEvent<byte[], byte[]>> changeEventConsumer,
+				Builder<ChangeEvent<byte[], byte[]>> debeziumEngineBuilder) {
+			return new EmbeddedEngineExecutorService(debeziumEngineBuilder.notifying(changeEventConsumer).build());
 		}
 
 		@Bean
--- a/applications/source/debezium-source/README.adoc
+++ b/applications/source/debezium-source/README.adoc
@@ -20,8 +20,9 @@ The `Debezium Source` currently supports CDC for multiple datastores: https://de
 //tag::configuration-properties[]
 $$debezium.copy-headers$$:: $$Copy Change Event headers into Message headers.$$ *($$Boolean$$, default: `$$true$$`)*
 $$debezium.debezium-native-configuration$$:: $$<documentation missing>$$ *($$Properties$$, default: `$$<none>$$`)*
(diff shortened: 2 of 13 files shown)


artembilan (approved):
LGTM, but cannot merge yet until current release is done.

functions/supplier/debezium-supplier/src/test/java/org/springframework/cloud/fn/supplier/debezium/custom/DebeziumEngineBuilderAutoConfigurationTest.java
```diff
@@ -57,8 +57,8 @@
  */
 @Tag("integration")
 @Testcontainers
-public class DebeziumEngineAutoConfigurationTest {
-	private static final Log logger = LogFactory.getLog(DebeziumEngineAutoConfigurationTest.class);
+public class DebeziumEngineBuilderAutoConfigurationTest {
```
onobc, line 60: Is there any reason that the auto-config test is in a different "custom" subpackage than the auto-config?

functions/supplier/debezium-supplier/src/test/java/org/springframework/cloud/fn/supplier/debezium/custom/DebeziumEngineBuilderAutoConfigurationTest.java
```diff
@@ -57,8 +57,8 @@
...
+	private static final Log logger = LogFactory.getLog(DebeziumEngineBuilderAutoConfigurationTest.class);
```
onobc, line 61: Switching to `LogAccessor` has some benefits, such as it auto-checking if the logging is at the requested level.

functions/supplier/debezium-supplier/src/test/java/org/springframework/cloud/fn/supplier/debezium/custom/DebeziumEngineBuilderAutoConfigurationTest.java
onobc, line 60: I think naming this "AutoConfigurationTest" is not accurate as it is an integration test of the functionality. 

I think we should keep this test, rename it `DebeziumEngineBuilderIntegrationTests`, create another `DebeziumEngineBuilderAutoConfgurationTests` that verifies autoconfig features such as the following:

- `@ConditionalOnMissingBean` beans can be customized (aka `custom<xyz>IsRespected()`) for the : offsetCommitPolicy, debeziumClock, completionCallback, connectorCallback ([example](https://github.com/spring-projects/spring-pulsar/blob/12e557c79ad7c923d538a044dfe67297ead167da/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java#L131))
- the `debeziumEngineBuilder` is configured using the passed in properties
- and also important is the `OnDebeziumConnectorCondition` where we can easily use the `ApplicationContextRunner.withFilteredClassLoader` to verify the condition

On one hand it might seem like we are testing simple things that should just work. Yes, but it is more about putting that in place so when someone decides to refactor later and then the "custom xyz" is no longer respected and user finds the bug. Similar to how some unit tests seem a waste at the beginning, but they end up paying off. I prefer to treat the auto-config in this same way. You will find the same treatment in the boot repo for the auto-configs that they support.

onobc (changes requested):
Looks good @tzolov - some minor adjustments requested.

onobc (changes requested):
Looks good @tzolov - some minor adjustments requested.

onobc (changes requested):
Looks good @tzolov - some minor adjustments requested.

onobc (changes requested):
Looks good @tzolov - some minor adjustments requested.

functions/supplier/debezium-supplier/src/test/java/org/springframework/cloud/fn/supplier/debezium/custom/DebeziumEngineBuilderAutoConfigurationTest.java
tzolov: Yes, because it shows how to implement a `custom` Consumer and `custom` ExecutionService instead of those implemented in the `DebeziumReactiveConsumerConfiguration`.  But i'm open for other package name suggestions.

onobc: That sounds more like a sample? The standard location for tests is in the same package as the unit it is testing - i would be in favor of sticking w/ that for the tests.

onobc:
> Looks good @tzolov - some minor adjustments requested.

Looks like Github went haywire on the comment LOL.

functions/supplier/debezium-supplier/src/test/java/org/springframework/cloud/fn/supplier/debezium/custom/DebeziumEngineBuilderAutoConfigurationTest.java
onobc: Ha, @artembilan just pointed out to me that this my comment was in the test code (yeh, I don't care which logger we use here and really not typical to have a logger in the test code).

PR #460, opened 2023-05-30
Debezium version update and new mssql it test

tzolov:
- Update debezium versin from 2.2.0.Final to 2.2.1.Final.
 - Add SQL Server IT.


applications/source/debezium-source/src/test/java/org/springframework/cloud/stream/app/source/debezium/databases/DebeziumDatabasesIntegrationTest.java
```diff
@@ -155,6 +157,49 @@ public void postgres() {
...
+
+				List<Message<?>> allMessages = new ArrayList<>();
+				Awaitility.await().atMost(Duration.ofMinutes(5)).until(() -> {
+					List<Message<?>> messageChunk = DebeziumTestUtils.receiveAll(outputDestination);
+					if (!CollectionUtils.isEmpty(messageChunk)) {
+						logger.info("Chunk size: " + messageChunk.size());
+						allMessages.addAll(messageChunk);
+					}
+
+					System.out.println("SIZE: " + allMessages.size());
```
onobc, line 193: Any reason to not use logger here OR to use sout above?

applications/source/debezium-source/src/test/resources/docker/mssql/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/mssql/server:2022-latest
```
onobc, line 1: Why can't we use Testcontainers for this? 

Also, there is some existing "sqlserver" directory under test/resources I believe.

applications/source/debezium-source/src/test/resources/docker/mssql/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+ENV ACCEPT_EULA y
+ENV MSSQL_SA_PASSWORD MyFancyPassword123
+ENV MSSQL_PID Standard
+ENV MSSQL_AGENT_ENABLED true
+# COPY ./init.sql .
+COPY ./init-with-signalling.sql ./init.sql
+COPY ./entrypoint.sh .
```
artembilan, line 8: Yeah... Looks like `ContainerState` has respective methods:
```
    /**
     *
     * Copies a file to the container.
     *
     * @param transferable file which is copied into the container
     * @param containerPath destination path inside the container
     */
    @SneakyThrows(IOException.class)
    default void copyFileToContainer(Transferable transferable, String containerPath) {
```

artembilan (changes requested):

applications/source/debezium-source/src/test/resources/docker/mssql/Dockerfile
tzolov: We need to instrument the SQL Server for [CDC](https://learn.microsoft.com/en-us/sql/relational-databases/track-changes/about-change-data-capture-sql-server?view=sql-server-ver16).  
Also, due to license rules we cannot distribute our own modified sql server image. Instead we have to build it on the fly. 

> Also, there is some existing "sqlserver" directory under test/resources I believe.
this old and will be removed.

applications/source/debezium-source/src/test/resources/docker/mssql/init-with-signalling.sql
```diff
@@ -0,0 +1,93 @@
+-- Create the test database
+CREATE DATABASE testDB;
```
onobc, line 2: This file gets renamed during dockerfile copy to init.sql, correct? If so, maybe we can remove the other `init.sql` that was copied over as it seems to be unused. Then just name this `init.sql`?

applications/source/debezium-source/src/test/resources/docker/mssql/Dockerfile
onobc: > Also, due to license rules we cannot distribute our own modified sql server image. Instead we have to build it on the fly.

I think we could use the `acceptLicense` API on the provided `MSSQLServerContainer` for this? 

It seems like we could do something like this instead:

```java
new MSSQLServerContainer("microsoft/mssql-server-linux:2017-CU9-GDR2")
				.acceptLicense()
				.withClasspathResourceMapping()
				.withCopyFileToContainer()
```
and get ride of the entrypoint and Dockerfile?

artembilan (approved):

onobc (approved):
@tzolov thanks for all of the updates. Nice work!

I just added a tiny-nit comment but this is approved w/ or w/o that being addressed.

Issue #441, opened 2023-03-16
Aggregator app encodes messages in base64


CEDDM:
I'm trying to use the `aggregator processor 2021.1.x` with `SCDF 2.9.6` and Kafka.

The app does aggregate the messages except I get a list of base64 encoded messages instead of json messages.

The splitted messages have all the contentType header set to "application/json".

Here is an example :  
The DSL : `test-aggregator = http | splitter | aggregator | log`  
Deployed with these properties :  

    version.http=3.2.1
    version.splitter=3.2.1
    version.aggregator=2021.1.x
    version.log=3.2.1
    
    app.http.server.port=8181
    app.splitter.splitter.expression=#jsonPath(payload, '$.store.book')
    app.aggregator.spring.cloud.stream.kafka.default.consumer.standard-headers=both

Then I post this JSON file (test.json) to the `http source` :  

    { "store": {
        "book": [
            {
                "author": "Nigel Rees",
                "title": "Sayings of the Century"
            },
            {
                "author": "Evelyn Waugh",
                "title": "Sword of Honour"
            },
            {
                "author": "Herman Melville",
                "title": "Moby Dick"
            },
            {
                "author": "J. R. R. Tolkien",
                "title": "The Lord of the Rings"
            }
        ]
    }}

To do that, I use the SCDF shell like this :

    http post --target http://<ip-http-source>:8181 --file test.json --contentType "application/json; charset=utf-8"

When I check the Kafka messages using Kowl after the splitter, I have the 4 books messages as JSON with the good `contentType` header. But after the aggregator, here is the results in the log sink :  

    [
      "eyJhdXRob3IiOiJOaWdlbCBSZWVzIiwidGl0bGUiOiJTYXlpbmdzIG9mIHRoZSBDZW50dXJ5In0=",
      "eyJhdXRob3IiOiJFdmVseW4gV2F1Z2giLCJ0aXRsZSI6IlN3b3JkIG9mIEhvbm91ciJ9",
      "eyJhdXRob3IiOiJIZXJtYW4gTWVsdmlsbGUiLCJ0aXRsZSI6Ik1vYnkgRGljayJ9",
      "eyJhdXRob3IiOiJKLiBSLiBSLiBUb2xraWVuIiwidGl0bGUiOiJUaGUgTG9yZCBvZiB0aGUgUmluZ3MifQ=="
    ]

I was expecting to get JSON messages like this : 
```
[
    {
        "author": "Nigel Rees",
[...]

PR #461, opened 2023-05-30
GH-441: Add `JsonBytesToMap` function

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/441

* Introduce a `JsonBytesToMap` as a part of a `payload-converter-function` module which is auto-discovered by Spring Cloud Function scanning algorithm - the `functions` package.
* Add a `payload-converter-function` as dependency into an `aggregator-function`
* Compose `jsonBytesToMap|aggregatorFunction` for the `aggregator-processor`
* Verify a `JsonBytesToMap` function in action with an `AggregatorProcessorTests`
* Mentioned such a payload conversion in the `aggregator-processor` README

### DIFF
--- a/applications/processor/aggregator-processor/README.adoc
+++ b/applications/processor/aggregator-processor/README.adoc
@@ -9,6 +9,10 @@ Change kafka to rabbit if you want to run it against RabbitMQ.
 
 === Payload
 
+If an input payload is a `byte[]` and content-type header is a JSON, then `JsonBytesToMap` function tries to deserialize this payload to a `Map` for better data representation on the output of the aggregator function.
+Also, such a `Map` data representation makes it easy to access to the payload content from SpEL expressions mentioned below.
+Otherwise, the input payload is left as is -  and it is the target application configuration to convert it into a desired form.
+
 == Options
 
 //tag::configuration-properties[]
--- a/applications/processor/aggregator-processor/README.adoc
+++ b/applications/processor/aggregator-processor/README.adoc
@@ -11,7 +11,7 @@ Change kafka to rabbit if you want to run it against RabbitMQ.
 
 If an input payload is a `byte[]` and content-type header is a JSON, then `JsonBytesToMap` function tries to deserialize this payload to a `Map` for better data representation on the output of the aggregator function.
 Also, such a `Map` data representation makes it easy to access to the payload content from SpEL expressions mentioned below.
-Otherwise, the input payload is left as is -  and it is the target application configuration to convert it into a desired form.
+Otherwise(including a deserialization error), the input payload is left as is - and it is the target application configuration to convert it into a desired form.
 
 == Options
 
--- a/applications/processor/aggregator-processor/src/test/java/org/springframework/cloud/fn/aggregator/AggregatorProcessorTests.java
+++ b/applications/processor/aggregator-processor/src/test/java/org/springframework/cloud/fn/aggregator/AggregatorProcessorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 the original author or authors.
+ * Copyright 2020-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


functions/function/payload-converter-function/src/main/java/functions/JsonBytesToMap.java
```diff
@@ -0,0 +1,75 @@
...
+
+		return message;
+	}
+
+	private Object payloadToMapIfCan(byte[] payload) {
+		try {
+			return this.objectMapper.readValue(payload, Map.class);
+		}
+		catch (IOException ex) {
+			// Was not able to construct the map from byte[] -- returning as is
```
onobc, line 70: Might be helpful to add a TRACE logging to allow debugging "Why is my payload not coming across as JSON map?" issues in the future.

applications/processor/aggregator-processor/README.adoc
```diff
@@ -9,6 +9,10 @@ Change kafka to rabbit if you want to run it against RabbitMQ.
 
 === Payload
 
+If an input payload is a `byte[]` and content-type header is a JSON, then `JsonBytesToMap` function tries to deserialize this payload to a `Map` for better data representation on the output of the aggregator function.
+Also, such a `Map` data representation makes it easy to access to the payload content from SpEL expressions mentioned below.
+Otherwise, the input payload is left as is -  and it is the target application configuration to convert it into a desired form.
```
onobc, line 14: **[TinyNit]** Nice doc! there is an extra space after the `-` on line 14. Also, I think it may be implied w/ "Otherwise" but might be helpful to explicitly add if the deserialize fails the input payload is left as is.

applications/processor/aggregator-processor/src/test/java/org/springframework/cloud/fn/aggregator/AggregatorProcessorTests.java
```diff
@@ -34,41 +39,60 @@
...
+
+			Person person1 = new Person("First1 Last1", "St. #1");
+			processorInput.send(createMessage(objectMapper.writeValueAsBytes(person1), 2, 2));
+
+			Person person2 = new Person("First2 Last2", "St. #2");
+			processorInput.send(createMessage(objectMapper.writeValueAsBytes(person2), 1, 2));
+
+			receive = processorOutput.receive(10_000, "jsonBytesToMapaggregatorFunction-out-0");
+
+			assertThat(receive).isNotNull();
```
onobc, line 73: Nice test addition! 

Wdyt about adding a check on that the `receive.payload` is instance of map? You do indirectly verify that via the `readValue` on 75.

applications/processor/aggregator-processor/src/test/java/org/springframework/cloud/fn/aggregator/AggregatorProcessorTests.java
```diff
@@ -34,41 +39,60 @@
...
+			List<Person> result =
+					objectMapper.readValue(receive.getPayload(),
+							objectMapper.constructType(new TypeReference<List<Person>>() { }));
+
+			assertThat(result).hasSize(2).contains(person1, person2);
```
onobc, line 78: **[TIP]** `assertThat(LIST).containsExactly(..)` is nice for this as it check both size and elements

applications/processor/aggregator-processor/src/test/java/org/springframework/cloud/fn/aggregator/AggregatorProcessorTests.java
artembilan: It is not a `Map`: the result of the `processorOutput.receive()` is a `byte[]`. That's why we have to deserialize it using an `ObjectMapper` and try to do that against expected types.

I cannot come up with an idea how to verify that input is a `Map` without a breaking use-case where it is really not a `Map`...

artembilan: Resolved.
See a fresh commit.

onobc (approved):
LGTM! Thanks Artem.

PR #463, opened 2023-05-31
Factor out debeiuzm builder into a function common autoconfig project

### LINKED ISSUE #462: Factor out the Debezium builder auto-configuration into its own function common project
Move the `DebeziumEngineBuilderAutoConfiguration` auto-configuration out of the debezium-supplier to a dedicated function common project.
Later would allow to reuse the auto-configuration for different customer use-cases, such as the debezium-supplier a Spring Integration debezium adapter, Spring Data cache replications ...

tzolov:
Move debezium builder into standalone function common project
    
     - Move the DebeziumEngineBuilderAutoConfiguration and META-INF configurations to new function project: debezium-autoconfigure.
     - Introduce the EmbeddedEngineExecutorService helper class for help DebezuimEngine implementations.
     - Move the relevat tests from debezium-supplier to the debeizum-autoconfigure.
     - Add detailer README documentation.
     - Adjust the debeizum-supplier and debezium-source configurations.
     - Clean the related tests.
     - 
  Resolves #462


functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,252 @@
...
+----
+[source, groovy, subs="normal", role="secondary"]
+.Gradle
+----
+compile "io.debezium:debezium-connector-postgres:{debezium-version}"
+----
+====
+
+
+Then implement your own `java.util.Consumer<ChangeEvent>` (or `ChangeConsumer<ChangeEvent>`) handler bean and wire it into a DebeziumEngine instance. For example:
```
artembilan, line 52: I don't think it is really a strict requirement to have such a consumer as a bean.
There is just enough to have a `DebeziumEngine.Builder` autowired, set some consumer into it and build an engine.
There is even no requirement to have that engine as a bean as well.
There is probably just enough to have something like `MyDebeziumService` marked with a `@Service` implementing a `SmartLifecycle`, and having that auto-configured `DebeziumEngine.Builder` injected.

Sure! It is possible what you show so far, but don't make end-users stuck just with only this style 😄

artembilan (changes requested):
I don't have too much objections with what I've seen here so far.
Just great, @tzolov ! 👍

functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumReactiveConsumerConfiguration.java
```diff
@@ -50,7 +51,7 @@
  * @author Artem Bilan
  */
 @Configuration
-@EnableConfigurationProperties(DebeziumProperties.class)
+@EnableConfigurationProperties({ DebeziumProperties.class, DebeziumSupplierProperties.class })
```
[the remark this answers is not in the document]
tzolov: Very good advice, thanks

functions/common/debezium-autoconfigure/README.adoc
tzolov: Good point. I've added some related notes under the sample. 
Maybe we can think of different or another sample to ilustrate it.

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -115,7 +114,7 @@ NOTE: The `EmbeddedEngineExecutorService` is optional and usually users would im
 For example the `DebeziumReactiveConsumerConfiguration` embeds an ExecutorService as part of the
 `Supplier<Flux<Message<?>>>` configuration.
 
-NOTE: Neither the `Consumer`/`ChangeConsumer` nor the `DebeziumEngine` need to be implemented as `@Bean`! It is just enough to have a `DebeziumEngine.Builder`, set some consumer into it, build an engine and run it from within some `@Service` implementation.
+NOTE: Neither the `Consumer`/`ChangeConsumer` nor the `DebeziumEngine` need to be `@Bean`! It is enough to set some consumer into it the `DebeziumEngine.Builder`, build an engine and run it from within some `@Service` implementation.
```
artembilan, line 117: Still not following `One sentence per line` rule...

functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumReactiveConsumerConfiguration.java
```diff
@@ -49,8 +51,8 @@
  * @author Christian Tzolov
  * @author Artem Bilan
  */
-@Configuration
-@EnableConfigurationProperties(DebeziumProperties.class)
+@AutoConfiguration(after = DebeziumEngineBuilderAutoConfiguration.class)
```
artembilan, line 54: Now we probably have to add `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports` for this project.

To be honest I think all of our out-of-the-box functions have to be auto-configuration. Otherwise their "drop-n-use" purpose is lost in target projects.

artembilan (changes requested):

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,280 @@
+# Debezium Auto-Configuration
+
+This module provides a generic https://debezium.io/documentation/reference/2.2/development/engine.html[DebeziumEngine.Builder] auto-configuration that can be reused and composed in other applications.
+
+NOTE: The `DebeziumEngine` doesn't not required Kafka or Kafka Connect as it runs embedded inside your application. This approach though comes with some delivery guarantee limitations as explained https://debezium.io/documentation/reference/2.2/development/engine.html#_handling_failures[here].
```
onobc, line 5: **[TinyNit]** `s/doesn't not/does not`

onobc, line 5: Should this be `IMPORTANT:` instead?

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,280 @@
...
+
+The `Debezium Engine` is a https://en.wikipedia.org/wiki/Change_data_capture[Change Data Capture] (CDC) utility, that allows *capturing* database change events and process them with custom `java.util.Consumer` or `io.debezium.engine.ChangeConsumer` event handler implementations.
```
onobc, line 7: Why is everything below here now italicized? I think the `#` in the link above kicked that on.

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,280 @@
...
+
+The `DebeziumEngine.Builder` auto-configuration is activated only if a https://debezium.io/documentation/reference/stable/connectors/index.html[Debezium Connector] is found on the classpath and the `debezium.properties.connector.class` property is set to point to that connector class.
+
+## Quick Start
+
+To use of the `DebeziumEngine` one need to (1) add required <<dependencies>> to the project, (2) provide the necessary Debezium <<configuration-options>>, (3) implement custom <<changeEvent-handlers>> and register it with the engine and (4) build and submit the DebeziumEngine to an `Executor` or `ExecutorService` for execution.
```
onobc, line 13: **[TinyNit]** "To use ~of~ the `DebeziumEngine` one need**s** to..."

onobc, line 13: These turn out to be more than is easily digestible (for me anyways) and may ~be better suited in a numbered bullet list. Wdyt?~

Actually after reading further down I see you outline these steps very well.  I take back the above request and instead think you could omit the "steps" here and instead say something like :

> To use the `DebeziumEngine` one need to follow a few simple steps - illustrated in the following sections.

Or some verbiage like that. My point is that naming the 5 steps here is a bit overwhelming but pointing to down below where you break them out very well may be an easier read. Wdyt?

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,280 @@
...
+To process the incoming database change events you need to include the Debezium `auto-configuration` dependency to your project:
+
+====
+[source, xml, subs="normal", role="primary"]
+.Maven
+----
+<dependency>
+    <groupId>org.springframework.cloud.fn</groupId>
+    <artifactId>debezium-autoconfigure</artifactId>
+    <version>{project-version}</version>
```
onobc, line 27: I think this should be `${project.version}`.

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,280 @@
...
+compile "io.debezium:debezium-connector-postgres:{debezium-version}"
+----
+====
+[[changeEvent-handlers]]
+### ChangeEvent Handlers
+
+Then implement your own `java.util.Consumer<ChangeEvent>` (or `ChangeConsumer<ChangeEvent>`) handler and wire it into a DebeziumEngine instance.
+
+It is just enough to have a `DebeziumEngine.Builder`, set your consumer implementation into it, build an engine and run it from within some `@Service` or alike implementation.
+Here are is an oversimplified example:
```
onobc, line 63: **[TinyNit]** "Here ~are~ is an.."

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,280 @@
...
+[source, java]
+----
+public void simple(DebeziumEngine.Builder<ChangeEvent<byte[], byte[]>> builder) {
+	Executors.newSingleThreadExecutor().execute(builder // <3>
+			.notifying(changeEvent -> System.out.println(changeEvent.value())) // <1>
+			.build()); // <2>
+}
+----
+<1> Implement and register a change events consumer.
```
onobc, line 72: I like this linking the steps to the sample NICE!

functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumReactiveConsumerConfiguration.java
onobc: I see now that I have gone through all commits that we have added the imports file and this is in fact an AC now! Yay! 

@tzolov do you mind exploring removing the config class from the source app pom.xml?

onobc (changes requested):
Awesome work @tzolov (as usual!). 

The docs you have added are excellent. 

I have minor comments to address but other than that - this one is definitely LGTM.

functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumReactiveConsumerConfiguration.java
tzolov: @onobc I added the META-INFO spring auto-config but I'm afraid i can not remove the config class configuration form the source app at the moment. 
This is an uncharted territory and the `app.configClass` is a compulsory argument for the `spring-cloud-dataflow-apps-generator-plugin`: 
```
Execution app-gen of goal org.springframework.cloud:spring-cloud-dataflow-apps-generator-plugin:1.0.7:generate-app failed: No key, method or field with name 'app.configClass' on line 25 
```
We have to raise and solve the issue in the plugin first to enable the auto-configuration.

functions/common/debezium-autoconfigure/README.adoc
tzolov: I wonder if we can inject variables with the real versions here

functions/common/debezium-autoconfigure/README.adoc
artembilan: I think this is just this text render bug. If you go to File View, everything looks OK.

functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumReactiveConsumerConfiguration.java
artembilan: That's fully different story.
If an auto-config was this function works even if it is imported again by that source application, we should leave that part outside of this PR and revise it when we already merge that plugin back to this project.
And probably already after we move functions into their own project.
There we would revise all the functions to be auto-configured since that would be a great feature in non-stream apps case where we want to compose functions ourselves.

artembilan (changes requested):

functions/supplier/debezium-supplier/src/main/java/org/springframework/cloud/fn/supplier/debezium/DebeziumReactiveConsumerConfiguration.java
onobc: I thought that may happen @tzolov . No worries. I agree w/ @artembilan on the timeline - no rush and it does not hurt anything to be in both places for now.

functions/common/debezium-autoconfigure/README.adoc
onobc: We could always add this in subsequent PR

artembilan: This is not possible.
See: this is a top-level project `README.adoc` which is fully static, unlike all those variants you suggest for docs generation.

functions/common/debezium-autoconfigure/README.adoc
tzolov: Escapted the # in engine.html$$#$$_handling_failures 
Hopefully this will help.

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,305 @@
+# Debezium Auto-Configuration
```
artembilan, line 1: This might cause the italic problem.
This is not MarkDown, but ASCIIDoc, so probably `=` should be used for head lines.
Although it renders in the view OK 🤷

artembilan (changes requested):

artembilan (changes requested):

onobc (approved):
Thanks for the updates @tzolov - LGTM

tzolov:
LGTM to me too ;)

artembilan (changes requested):

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,322 @@
+= Debezium Auto-Configuration
+
+This module provides a generic https://debezium.io/documentation/reference/2.2/development/engine.html[DebeziumEngine.Builder] auto-configuration that can be reused and composed in other applications.
+
+IMPORTANT: The `DebeziumEngine` does not required Kafka or Kafka Connect as it runs embedded inside your application.
+This approach though comes with some delivery guarantee limitations as explained https://debezium.io/documentation/reference/2.2/development/engine.html$$#$$_handling_failures[here].
```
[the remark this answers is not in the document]
onobc: Good catch - many comments on this PR and I missed this one. Looks like you are the final arbiter on the PR now ;)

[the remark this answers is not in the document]
artembilan:
```suggestion
This approach though comes with some delivery guarantee limitations as explained https://debezium.io/documentation/reference/2.2/development/engine.html#\_handling_failures[here].
```

functions/common/debezium-autoconfigure/README.adoc
```diff
@@ -0,0 +1,353 @@
+= Debezium Auto-Configuration
+
+This module provides a generic https://debezium.io/documentation/reference/2.2/development/engine.html[DebeziumEngine.Builder] auto-configuration that can be reused and composed in other applications.
+
+IMPORTANT: The `DebeziumEngine` does not required Kafka or Kafka Connect as it runs embedded inside your application.
+This approach though comes with some delivery guarantee limitations as explained https://debezium.io/documentation/reference/2.2/development/engine.html#%5Fhandling_failures[here].
```
artembilan, line 6: @onobc ,

is it better now from an *italic* perspective?

artembilan (approved):

PR #464, opened 2023-06-04
Remove legacy config properties metadata files

### LINKED ISSUE #252: Remove legacy config properties metadata files
We should be able to at, at some point, delete the legacy config metadata files. They are currently kept around simply to not break older installs that may still be referencing these filenames. 

These legacy files contain non-inclusive terms "whitelist". ([example](https://github.com/spring-cloud/stream-applications/blob/main/applications/sink/log-sink/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties)).

We first need to look at the date the change was made (released). And then figure out when the "support" for the previous versions containing the old files is expired to figure out our earliest date we can pull the trigger on this. 

One thought is to do it when we do the Boot3 based release of the apps (aka the 2022 Dec timeframe)

onobc:
This commit removes all legacy config props metadata files named `dataflow-configuration-metadata-whitelist.properties`.

This is safe with respect to SCDF because stream apps 4.x require SCDF 2.11.x to launch the apps. Therefore, no existing SCDF customers on an older SCDF version (that only supports the legacy file) will be impacted - the users will all be on the latest version of SCDF.

Fixes #252


applications/sink/rsocket-sink/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
```diff
@@ -1,2 +1 @@
 configuration-properties.classes=org.springframework.cloud.fn.consumer.rsocket.RsocketConsumerProperties
-
```
artembilan, line 2: I don't understand: you removed all others, but then you just fix a new line in this one.
Why don't remove this file as well?

applications/sink/rsocket-sink/README.adoc
```diff
@@ -8,10 +8,6 @@ RSocket sink to send data using RSocket protocols' fire and forget strategy.
 The **$$rsocket$$** $$sink$$ has the following options:
 
 //tag::configuration-properties[]
-$$rsocket.consumer.host$$:: $$RSocket host.$$ *($$String$$, default: `$$localhost$$`)*
-$$rsocket.consumer.port$$:: $$RSocket port.$$ *($$Integer$$, default: `$$7000$$`)*
-$$rsocket.consumer.route$$:: $$Route used for RSocket.$$ *($$String$$, default: `$$<none>$$`)*
-$$rsocket.consumer.uri$$:: $$URI that can be used for websocket based transport.$$ *($$URI$$, default: `$$<none>$$`)*
```
artembilan, line 14: Is it really correct to remove this props?
What happened? Why don't you remove similar then for all other READMEs?

artembilan (changes requested):

applications/sink/rsocket-sink/README.adoc
onobc: No. I needed to re-run the build after adding back in the file that was accidentally removed (https://github.com/spring-cloud/stream-applications/pull/464#discussion_r1218388487).

I have the changes locally and forgot to push the subsequent commit (doing that now).

applications/sink/rsocket-sink/src/main/resources/META-INF/dataflow-configuration-metadata-whitelist.properties
artembilan: That means that there has to be a new file in the RSsocket Sink module instead of this one, like you did for PG Copy.
But we still have discussion with you on a wrong file and probably talk about different things.
Am I missing anything else as a purpose of this PR?

onobc: Yeh, I got confused on the files. Good catch @artembilan (aka Eagle Eye). It is updated now (stripped the legacy filename ending).

artembilan (approved):

artembilan:
Will merge when CI build is green.

Issue #265, opened 2022-05-10
Add Kafka source and sink apps


onobc:
To complete parity w/ the Rabbit source / sink, lets add a Kafka source/sink app. 

Motivation: https://github.com/spring-cloud/stream-applications/issues/202

Zishnu:
Please add the external kafka source and sink requirements as in enterprise application sceneries. The source and target application would be writing messages to external broker which is different from message broker topic.

artembilan:
Reopened for Supplier and Source implementation part.

fengnv17:
> Reopened for Supplier and Source implementation part.
could you tell the date for kafka supplier? thanks.

artembilan:
> could you tell the date for kafka supplier? thanks.

I'm looking into that these days.
So, we may have a SNAPSHOT for `kafka-supplier`  next week.
If you are asking about GA release, then ti might be something like a month yet.

PR #469, opened 2023-06-07
GH-265: Add Kafka Publisher and Sink

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/265

* Make `kafkaConsumer` based on the Spring for Apache Kafka auto-configuration and Spring Integration channel adapter, essentially `KafkaProducerMessageHandler`
* Make this `KafkaConsumerConfiguration` as an auto-configuration by itself.
* Expose those simple properties required by the `KafkaProducerMessageHandlerSpec`
* Add `kafka-sink` module based on the `kafka-consumer`


artembilan:
I would prefer to have `supplier` and `source` as an individual PR: too many things are involved in the change.

sobychacko:
Took a quick pass-through. That looks good, @artembilan. As you indicated in a comment, Kafka Consumer, in this context, is very confusing as this is doing exactly the opposite (i.e., producing to Kafka :-) ). But I don't see an easy solution other than documenting it.

artembilan:
Thanks, @sobychacko !

I wonder if I still need to add this new app into some configs.
For example I see properties like this `kafka-apps-maven.properties`, `rabbit-apps-maven-repo-url.properties` etc. in the `stream-applications-descriptor`.
Not sure what are those and if they are generated somehow automatically or so.

onobc:
> ** wonder if I still need to add this new app into some configs.

@artembilan take a peek at these: https://github.com/spring-cloud/stream-applications/blob/main/stream-applications-release-train/stream-applications-descriptor/src/main/resources/META-INF/kafka-apps-maven.properties

onobc (changes requested):
@artembilan great work!

I have no complaints other than the overloaded "Kafka Consumer" you and @sobychacko already mentioned. Here are my thoughts on that:

I am fairly certain the KafkaConsumerConfiguration and `kafka.consumer` property prefix will be SUPER SUPER confusing to Spring Kafka users.

I would be in favor of including something in the name to differentiate it, wdyt about `KafkaConsumerFunctionConfiguration` and `kafka.consumer-function`?

artembilan:
OK. How about to name it as `kafka-publisher` and so on for all the configurations and bean names?
In the end it doesn't matter how you call it here since the contract is still fulfilled - `Consumer<Message<?>> kafkaPublisher`.
Or `producer`, or `sender`...

sobychacko:
I like `kafka-publisher`, even though it might be going away from our established conventions. It takes all the confusion away as to what this module does.

functions/consumer/kafka-publisher/src/main/java/org/springframework/cloud/fn/consumer/kafka/KafkaPublisherConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+
+	@Bean
+	public KafkaProducerMessageHandler<?, ?> kafkaProducerMessageHandlerSpec(KafkaTemplate<?, ?> kafkaTemplate,
+			KafkaPublisherProperties kafkaPublisherProperties,
+			PublishSubscribeChannel kafkaPublisherSuccessChannel,
+			PublishSubscribeChannel kafkaPublisherFailureChannel,
+			PublishSubscribeChannel kafkaPublisherFuturesChannel,
+			@Nullable ComponentCustomizer<KafkaProducerMessageHandlerSpec<?, ?, ?>> kafkaProducerSpecComponentCustomizer) {
+
+		var kafkaProducerMessageHandlerSpec = Kafka.outboundChannelAdapter(kafkaTemplate);
```
onobc, line 69: I notice you start to use `var` here but then do not use it in below or in the tests? Intentional? I have switched to use it everywhere I can and it has made my DX a bit better.

onobc (approved):

functions/consumer/kafka-publisher/src/main/java/org/springframework/cloud/fn/consumer/kafka/KafkaPublisherConfiguration.java
artembilan: I don't know what is DX, but I use `var` only in those cases where the line for variable definition is too long 😄

onobc: Ahh "DX == developer experience" (ie it makes me a bit happier when I use this, my developer experience increases). 

I have found that `var` does clean up the code a bit.

artembilan: Well, I use whatever my IDEA generates for me when I type `.var` suffix after expression.
And put `var` instead of type, only of it is log.
If my IDEA will generate `var` after that `.var`, I'll be OK with the result 😸

Issue #470, opened 2023-06-08
Update to Spring Boot 3.1.x 


onobc:
Update to latest Spring Boot 3.1.x so we can GA w/ latest rather than SB 3.0.x.

Results from the update will be reported in this issue.

onobc:
Closed via https://github.com/spring-cloud/stream-applications/commit/d73a8f68416d7c08d1c474c38b4a00b4e26a77d9

PR #473, opened 2023-06-09
Update to Spring Boot 3.1.0

onobc:
See #470


artembilan (changes requested):
Shouldn't this one be against `boot_3_1` branch?

PR #474, opened 2023-06-09
GH-265: Add Kafka Supplier & Source

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/265


applications/source/kafka-source/README.adoc
```diff
@@ -0,0 +1,29 @@
...
+$$ack-discarded$$:: $$Whether to acknowledge discarded records after 'RecordFilterStrategy'.$$ *($$Boolean$$, default: `$$false$$`)*
+$$topic-pattern$$:: $$Apache Kafka topics pattern to subscribe.$$ *($$Pattern$$, default: `$$<none>$$`)*
+$$topics$$:: $$Apache Kafka topics to subscribe.$$ *($$String[]$$, default: `$$<none>$$`)*
+
+=== spring.kafka
+
+$$bootstrap-servers$$:: $$Comma-delimited list of host:port pairs to use for establishing the initial connections to the Kafka cluster. Applies to all components unless overridden.$$ *($$List<String>$$, default: `$$<none>$$`)*
+$$client-id$$:: $$ID to pass to the server when making requests. Used for server-side logging.$$ *($$String$$, default: `$$<none>$$`)*
+$$properties$$:: $$Additional properties, common to producers and consumers, used to configure the client.$$ *($$Map<String, String>$$, default: `$$<none>$$`)*
+//end::configuration-properties[]
```
artembilan, line 27: Pay attention, please, to this content from a `KafkaProperties`.
It is definitely not the whole list of props presented over there.
Not sure what is going on, but probably some issue in the doc generation plugin.

functions/supplier/kafka-supplier/README.adoc
```diff
@@ -0,0 +1,28 @@
+# Apache Kafka Consumer Supplier
+
+A `Consumer<Message<?>>` that allows to publish messages to Apache Kafka topic.
+
+
+## Beans for injection
+
+The `KafkaSupplierConfiguration` is an auto-configuration, so no need to import anything.
+
+The `Supplier<Flux<Message<?>>> kafkaSupplier` bean can be injection into target service for consuming data from Kafka topics.
```
onobc, line 10: `s/injection/injected`

functions/supplier/kafka-supplier/README.adoc
```diff
@@ -0,0 +1,28 @@
+# Apache Kafka Consumer Supplier
+
+A `Consumer<Message<?>>` that allows to publish messages to Apache Kafka topic.
```
onobc, line 3: I believe this was copy/pasta from the `kafka-publisher` as it does not talk of supplier (but rather consumer).

onobc (changes requested):
Great work @artembilan , I have a couple of minor comments to address and then I also gave some thoughts on the missing config props in the readme. Other than that, LGTM.

applications/source/kafka-source/README.adoc
artembilan: I don't think that Christian's changes in the apps plugin have any effect on a docs plugin.
I think the way  have it fixed now with nested classes in that our properties file is correct.

artembilan:
Now I've noticed that we have a top-level https://github.com/spring-cloud/stream-applications/blob/main/README.adoc with some tables pointing to our functions and apps.
What I fail to understand how they are organized.
Any pointers, please, at what row to inserts these new Kafka artifacts?
Thanks

applications/sink/kafka-sink/src/main/resources/META-INF/dataflow-configuration-metadata.properties
```diff
@@ -1,2 +1,3 @@
 configuration-properties.classes=org.springframework.cloud.fn.consumer.kafka.KafkaPublisherProperties, \
-  org.springframework.boot.autoconfigure.kafka.KafkaProperties
+  org.springframework.boot.autoconfigure.kafka.KafkaProperties, \
+org.springframework.boot.autoconfigure.kafka.KafkaProperties$Producer
```
onobc, line 3: Should we also expose the common properties they may use to configure connection to Kafka?
- KP.properties
- KP.ssl (can be done on KP.producer.security though)
- KP.security (can be done on KP.producer.security though)
- KP.retry

onobc, line 3: I wonder if there are other apps that use config props that have nesting and we just did not notice until this one?

artembilan: I tried that and it is got a bit messy: I cannot restrict it only to those props which I need here. With `org.springframework.boot.autoconfigure.kafka.KafkaProperties$Security` it shows also a `Consumer`, `Admin` and `Streams` parts.
If that is still a goal, I can go ahead and add all of them: even if we have some extra props, we would be covered for all those we really need.

onobc: I figured you probably explored this already. 

I have mixed feelings because KafkaProperties is so huge and many of the properties are unrelated to producing. Having more unrelated props documented may be more confusing for users than listing the producer-specific ones. I mean, all the other props will work - they just will not be advertised in the README. 

Based on that, I agree w/ the current changes.

onobc: Created https://github.com/spring-cloud/stream-applications/issues/475

onobc (approved):

Issue #475, opened 2023-06-12
Update apps metadata to include nested config props


onobc:
We recently discovered that some apps use config props that in turn has nested config props (eg `KafkaProperties$ProducerProperties` in `kafka-source`). 

The following apps need updating (there may be others as well):
- sftp-sink
- ftp-source
- ???

              I wonder if there are other apps that use config props that have nesting and we just did not notice until this one?

_Originally posted by @onobc in https://github.com/spring-cloud/stream-applications/pull/474#discussion_r1227068793_

artembilan:
> The following apps need updating (there may be others as well):

Sorry, that was not I meant.
Those apps have props like this:
```
configuration-properties.classes=org.springframework.cloud.fn.supplier.sftp.SftpSupplierProperties, \
  org.springframework.cloud.fn.supplier.sftp.SftpSupplierProperties$Factory, \
  org.springframework.cloud.fn.supplier.sftp.SftpSupplierProperties$SortSpec, \
  org.springframework.cloud.fn.common.file.FileConsumerProperties, \
  org.springframework.cloud.fn.common.metadata.store.MetadataStoreProperties, \
  org.springframework.cloud.fn.common.metadata.store.MetadataStoreProperties$Gemfire, \
  org.springframework.cloud.fn.common.metadata.store.MetadataStoreProperties$Redis, \
  org.springframework.cloud.fn.common.metadata.store.MetadataStoreProperties$DynamoDb, \
  org.springframework.cloud.fn.common.metadata.store.MetadataStoreProperties$Jdbc, \
  org.springframework.cloud.fn.common.metadata.store.MetadataStoreProperties$Zookeeper, \
  org.springframework.cloud.fn.common.metadata.store.MetadataStoreProperties$Mongo \
```

With that I mean the place where I took an idea how that can be provided.
And that was a pointer that we have apps with nested props.

I'll take this and will try to take care about missed parts if that.

PR #476, opened 2023-06-12
GH-475: Add some missed nested props for apps

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/475

* Fix JDBC props for limited listed number of them


onobc (approved):
Thanks for all of these updates @artembilan

PR #482, opened 2023-06-22
Upgrade debezium to 2.3.0.Final

tzolov:


applications/source/debezium-source/README.adoc
```diff
@@ -104,32 +104,32 @@ Those properties can be used by prefixing them by the `debezium.properties.` pre
...
 
-|https://debezium.io/documentation/reference/2.2/connectors/vitess.html[Vitess]
-|https://debezium.io/documentation/reference/2.2/connectors/vitess.html#vitess-connector-properties
+|https://debezium.io/documentation/reference/2.3/connectors/vitess.html[Vitess]
+|https://debezium.io/documentation/reference/2.3/connectors/vitess.html#vitess-connector-properties
 
-|https://debezium.io/documentation/reference/2.2/connectors/spanner.html[Spanner]
-|https://debezium.io/documentation/reference/2.2/connectors/spanner.html#spanner-connector-properties
+|https://debezium.io/documentation/reference/2.3/connectors/spanner.html[Spanner]
+|https://debezium.io/documentation/reference/2.3/connectors/spanner.html#spanner-connector-properties
```
artembilan, line 132: That's exactly what I meant about support burden if we don't use "versionless" links which essentially points to the latest versions anyway.
Like this one: https://debezium.io/documentation/reference/connectors/spanner.html#spanner-connector-properties

🤷

functions/supplier/debezium-supplier/pom.xml
```diff
@@ -15,7 +15,7 @@
     <description>Debezium Supplier</description>
 
     <properties>
-        <version.debezium>2.3.0.CR1</version.debezium>
+        <version.debezium>2.3.0.Final</version.debezium>
```
artembilan, line 18: I believe we can manage this version in a single place where we would just import `debezium-bom`

artembilan (changes requested):

functions/supplier/debezium-supplier/pom.xml
tzolov: Yeh, this is a recurring discussion. 
Currently the functions don't have a bom for external dependencies.  The existing `function-dependecies` is only providing a BOM for sharing internal dependencies between the `functions` and `applications`. 
So should we start crating a bom project for evry function common project that might need it or create one shared e.g. something like `function-common-dependecies` bom? 
For either decision is it right to do it at this release stage?

artembilan: Heads up! 
Turned out it is wrong to use `debezium-bom`.
They have too many not project related deps over there.
The point of BOM to manage only your own modules, but not the rest of transitive deps.
🤷

artembilan (changes requested):

PR #484, opened 2023-07-14
Upgrade to AWS SDK v2

artembilan:
* Upgrade to Spring Integration AWS `3.0.1`
* Upgrade to Spring Cloud AWS `3.0.1`
* Use `spring-cloud-aws-dependencies` BOM to manage Spring Cloud AWS deps, as well as AWS SDK
* Remove `AmazonS3Properties` in favor of similar properties in the Spring Cloud AWS auto-configuration
* Remove `CompatibleStorageAmazonS3Configuration` in favor of its code migration to the single `AmazonS3Configuration`
* Fix `supplier`, `consumer`, `source` and `sink` for S3 to use API and programing model from the mentioned upgrades
* Add extra `dataflow-configuration-metadata` entries for the mentioned properties from the Spring Cloud AWS auto-configuration
* Regenerate README for `s3-source` and `s3-sink`
* Fix `metadata-store-common` to use the latest Spring Integration AWS for DynamoDB store impl


artembilan:
This might easy justify an RC2 - too many breaking changes.

I wish I didn't forget to do this a month ago when I released Spring Integration AWS `3.0` 😢

artembilan:
And the question.

There is a module like `stream-applications-integration-tests`.
Do we use it any where?
Do we really need to update its integration tests to our latest changes?
Looks like that one rely on `latest` Docker images which I'm not sure when we make available.
Perhaps already after release...

Thanks

corneil (approved):

PR #495, opened 2023-09-05
Upgrade debezium to 2.3.3.Final

tzolov:


functions/common/debezium-autoconfigure/src/main/java/org/springframework/cloud/fn/common/debezium/DebeziumProperties.java
```diff
@@ -126,6 +126,7 @@ public void setOffsetCommitPolicy(DebeziumOffsetCommitPolicy offsetCommitPolicy)
 	public Properties getDebeziumNativeConfiguration() {
 		Properties outProps = new java.util.Properties();
 		outProps.putAll(this.getProperties());
+		System.out.println(outProps.toString());
```
artembilan, line 129: SOUT?
In production code?

artembilan (changes requested):

functions/common/debezium-autoconfigure/src/main/java/org/springframework/cloud/fn/common/debezium/DebeziumProperties.java
tzolov: cleaned!

Issue #500, opened 2023-09-13
Header serialization exception in aggregator processor with Redis


CEDDM:
When I configure my aggregator with `Redis` to cache JSON messages, I get a serialization exception because the object does not implement `Serializable`. This is an known problem with Redis default serializer (`JdkSerializationRedisSerializer`).

The problem is we don't have a real Java object when the aggregator is used in `SCDF` (the messages come from Kafka or RabbitMQ)
The recommanded workaround I found is to use `GenericJackson2JsonRedisSerializer` but there is no way to do this without forking the aggregator (see the answer of @artembilan here : https://stackoverflow.com/questions/77088203/redis-serializer-properties-in-scdf).

It would be great to add a property to set the Redis serializer in the app configuration.

EDIT : According to @artembilan analysis, the problem comes from headers not the payload so the description above is wrong. Some headers must be filtered if not serializable

artembilan:
Any explanation how an aggregator processor receives from the binder something what is not `byte[]`?
Well, I may understand that `JsonBytesToMap` function may have an effect since default content-type in Spring Cloud Stream is JSON.
But again: the result `Map` is `Serializable`.

May you can share with us a stack trace from that serialization error, so we would have some clue what is going on?

Thanks

CEDDM:
Sorry I forgot to mention that I don't use the last version of the aggregator. I use `2021.1.2` version.
I will give a try with the latest version containing the `JsonBytesToMap` evolution and come back to you.

CEDDM:
In fact I think the problem happens before the aggregation when the app is caching individual messages to `Redis` so it's not a `Map` at this point. And yes I don't understand why there is a serialization problem as the message is already `byte[]`...
Anyway I will try with the latest version and post the stacktrace

CEDDM:
I just see that `2021.1.2` is still the latest stable version so I couldn't try with a more recent one.
But my aggregation is :
`aggregation=#this.![@jacksonObjectMapper.readValue(payload, T(java.util.Map))]`
So I think it's pretty much the same as `JsonBytesToMap` 

Here is the stacktrace : 
```
[...]

PR #503, opened 2023-09-22
GH-500: Workaround for non-serializable header

artembilan:
Fixes https://github.com/spring-cloud/stream-applications/issues/500

When `listeners` are provided for `DefaultKafkaConsumerFactory`, the target `KafkaConsumer` instance is proxied.
The `java.lang.reflect.Proxy` is `Serializable`,
but the value it is wrapping is not.
When the `MessageHeaders` is serialized (e.g. into persistent `MessageStore`), it checks for `Serializable` type only on top-level object of the header. Therefore, the `Proxy` is passing condition, but eventually we fail with `NotSerializableException`, since the proxied object is not like that

* Remove `kafka_consumer` from a message before it reaches an aggregator with its logic to serialize message into the store

This is a workaround until Spring for Apache Kafka is released with the fix: https://github.com/spring-projects/spring-kafka/pull/2822


sobychacko (approved):

onobc (approved):
LGTM - nice and simple. Thanks @artembilan

PR #509, opened 2023-11-13
Fix scope of function-test-support dependency

cachescrubber:
### Fix scope of function-test-support dependency

I noticed unexpected logging behaviour when using `sftp-consumer` caused by the missing test scope of the `function-test-support` dependency.

Workaround:

```
    <dependency>
      <groupId>org.springframework.cloud.fn</groupId>
      <artifactId>sftp-consumer</artifactId>
      <version>4.0.1-SNAPSHOT</version>
      <exclusions>
        <exclusion>
          <groupId>org.springframework.cloud.fn</groupId>
          <artifactId>function-test-support</artifactId>
        </exclusion>
      </exclusions>
    </dependency>

```


onobc:
Thanks for the contribution @cachescrubber !

PR #544, opened 2024-06-02
Add spring.binders for SCSt test binder

onobc:
Works around issue in SCSt `4.1.2` where the spring.binders was removed from the 
`spring-cloud-stream-test-binder` module.

cc: @sobychacko


onobc:
I'm going to merge this in order to get a good main build.

PR #549, opened 2024-06-14
Add missing metadata jars to docker descriptors.

corneil:
Since the stream apps built with pack doesn't have the metadata in the same label we should add the metadata jars to the docker descriptors.


corneil:
builds do provide for dataflow-configuration.properties to produce the correct labels.

PR #554, opened 2024-09-03
Add Micrometer tracing

onobc:
This PR adds the required dependencies for Micrometer Tracing with the Brave bridge
with the ability to report trace to Zipkin or Wavefront. 

Also the following unrelated commits are piggybacking:
- covenience script to do a full build w/o tests
- update to Prometheus Rsocket 2.0.0-M1

> [!IMPORTANT]
> During merge please keep the commits separate as they are unrelated

### DIFF
--- a/applications/stream-applications-core/pom.xml
+++ b/applications/stream-applications-core/pom.xml
@@ -18,7 +18,7 @@
     <properties>
         <stream-apps-core.version>5.0.1-SNAPSHOT</stream-apps-core.version>
         <java-functions.version>5.0.1-SNAPSHOT</java-functions.version>
-		<prometheus-rsocket.version>1.5.2</prometheus-rsocket.version>
+        <prometheus-rsocket.version>2.0.0-M1</prometheus-rsocket.version>
         <spring-cloud-dataflow-apps-generator-plugin.version>1.0.14</spring-cloud-dataflow-apps-generator-plugin.version>
         <spring-cloud-dataflow-apps-docs-plugin.version>1.0.14</spring-cloud-dataflow-apps-docs-plugin.version>
         <spring-cloud-dataflow-apps-metadata-plugin.version>1.0.14</spring-cloud-dataflow-apps-metadata-plugin.version>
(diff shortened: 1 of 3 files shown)


applications/stream-applications-core/pom.xml
```diff
@@ -18,7 +18,7 @@
     <properties>
         <stream-apps-core.version>5.0.1-SNAPSHOT</stream-apps-core.version>
         <java-functions.version>5.0.1-SNAPSHOT</java-functions.version>
-		<prometheus-rsocket.version>1.5.2</prometheus-rsocket.version>
+        <prometheus-rsocket.version>2.0.0-M1</prometheus-rsocket.version>
```
artembilan, line 21: This would be possible if we would plan `5.1` version for this project.
Or `1.x` generation for that dependency is not compatible with current Micrometer?

applications/stream-applications-core/pom.xml
```diff
@@ -144,28 +144,37 @@
...
                                 <maven>
                                     <dependencyManagement>
+                                        <dependency>
+                                            <groupId>io.micrometer.prometheus</groupId>
+                                            <artifactId>prometheus-rsocket-spring</artifactId>
+                                            <version>${prometheus-rsocket.version}</version>
+                                        </dependency>
+                                        <dependency>
+                                            <groupId>io.micrometer.prometheus</groupId>
+                                            <artifactId>prometheus-rsocket-client</artifactId>
```
artembilan, line 175: Doesn't this come as a transitive dep of the previous one?

applications/stream-applications-core/pom.xml
```diff
@@ -214,17 +223,37 @@
...
-											<version>${prometheus-rsocket.version}</version>
-										</dependency>
                                         <dependency>
                                             <groupId>io.micrometer</groupId>
                                             <artifactId>micrometer-registry-wavefront</artifactId>
                                         </dependency>
+                                        <!-- Tracing -->
+                                        <dependency>
+                                            <groupId>io.micrometer</groupId>
+                                            <artifactId>micrometer-tracing-bridge-brave</artifactId>
```
artembilan, line 243: Wouldn't be there something like OTel support request from users?

artembilan (changes requested):

applications/stream-applications-core/pom.xml
onobc: Yep, good catch!

applications/stream-applications-core/pom.xml
onobc: Its possible. I would rather do that in a separate PR as this is to get feature parity w/ what was removed w/ Sleuth.

applications/stream-applications-core/pom.xml
onobc: The `1.x` generation is based on Spring Boot 2.7.x. The `2.x` generation is based on Spring Boot 3.3.x. I should have pointed this out w/ a self-review comment. Thanks for bringing it up.

artembilan: I know that was unfortunate that we had missed this before 5.0 GA, but to go on with such an upgrade I'd like to see a GA version as well. This might be just a dependency of our uber artifact (or Docker image), but still would be great to have it in Maven Central for consistency and so on.
Do you have any dates when this Prometheus Rsocket library is going 2. GA?

artembilan (changes requested):

applications/stream-applications-core/pom.xml
onobc: The GA is end of September last time I checked w/ Micrometer team. We are also using the M1 in Dataflow until that time.

onobc: This would stay in the snapshot until the GA is available.

applications/stream-applications-core/pom.xml
```diff
@@ -156,9 +156,11 @@
                                     <!-- Wavefront Tracing -->
                                     <wavefront.application.name>"${spring.cloud.dataflow.stream.name:unknown}</wavefront.application.name>
                                     <wavefront.application.service>${spring.cloud.dataflow.stream.app.label:unknown}-${spring.cloud.dataflow.stream.app.type:unknown}"</wavefront.application.service>
-                                    <!-- Zipkin Tracing -->
+                                    <!-- Disable Tracing and all Tracing exporters -->
                                     <management.tracing.enabled>false</management.tracing.enabled>
                                     <management.tracing.sampling.probability>1.0</management.tracing.sampling.probability>
+                                    <management.zipkin.tracing.enabled>false</management.zipkin.tracing.enabled>
```
onobc, line 162: Realized I missed this property. Piggy backed in since this PR still in motion.

artembilan:
@onobc ,

looks like something is missing yet: https://github.com/spring-cloud/stream-applications/actions/runs/10707057201.

Apparently the deps management must be not only in one place.

Thanks

onobc:
> @onobc ,
> 
> looks like something is missing yet: https://github.com/spring-cloud/stream-applications/actions/runs/10707057201.
> 
> Apparently the deps management must be not only in one place.
> 
> Thanks

Looking into it. I think the update I did w/ transitive dependency had a side effect as this was working locally. I will get it squared away and a fix submitted. Thanks for the review and merge.

PR #564, opened 2024-11-08
Update jfrog

corneil:
Update to JFrog action v4.


onobc (approved):
Thanks for the updates @corneil

Issue #550, opened 2024-07-01
Test and Java-CFEnv version


corneil:
There is a new major version of Java CF Env. Currently at 3.1.5
Spring Boot 3.x application requires java-cfenv 3.1.5 for CloudFoundry.

We need to test the new version with CF.

PR #567, opened 2024-11-08
Simplify CI-PR

corneil:
Removed checks for changed modules and ensure the build scripts works in the absence of credentials.

See #550


.github/workflows/ci-pr.yml
```diff
@@ -135,9 +66,23 @@ jobs:
       - name: 'Built'
         shell: bash
         run: echo "::info ::Built"
+  scan:
```
onobc, line 69: Let's remove this scan from the PR build as we do not accept dep. updates in PRs and Trivy is failing 99% of the time due to its rate limiting.

.github/workflows/ci-pr.yml
```diff
@@ -26,18 +26,10 @@ jobs:
...
-          java-version: 17
-      - name: 'Configure: Install GraalVM'
-        if: ${{ github.base_ref == 'main' }}
-        uses: graalvm/setup-graalvm@v1
-        with:
-          version: 'latest'
-          java-version: '17'
-          components: 'js'
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          java-version: '17.0.12'
```
onobc, line 31: Any particular reason so specific w/ patch version?

corneil: `setup-java@v4` requires complete version for some reason it doesn't select latest when you only specify `17`

PR #568, opened 2024-11-12
Testing setup-java version selection

corneil:
Test setup-java@v4 for selecting latest version of distribution.


corneil:
Tested multiple combinations to determine which work and which fails
setup-java@v4 with `java-version: 17` and `distribution: graalvm` fails while `java-version: 21` and `distribution: graalvm` works as well as `java-version: 17` and `distribution: liberica`

Issue #584, opened 2024-11-25
Update Spring Boot to 3.3.6


onobc:
Spring Boot needs to be updated to 3.3.6 before we can release main (5.0.1).

Also update all other dependencies to the latest patch versions.

PR #587, opened 2024-11-25
Update Spring Boot to 3.3.6

onobc:
This updates Spring Boot to version 3.3.6 and other various dependencies to their latest available patch version.

See #584


corneil:
@onobc I believe we can merge this so that main is sound and probably branch 5.0.x tomorrow.

onobc:
I believe this has already been done in https://github.com/spring-cloud/stream-applications/commit/e4231b0f4e59171e451e6e5e1a574ada759b7b65 @corneil

PR #590, opened 2024-12-02
Fix build-name / number from input.

corneil:
Updates JFROG_CLI_BUILD_NAME and JFROG_CLI_BUILD_NUMBER from input to workflow.


onobc (approved):
Thanks for the update @corneil - LGTM.

PR #608, opened 2025-01-17
Override Spring Cloud Function version

onobc:
This commit overrides the `spring-cloud-function` version manged by `spring-cloud-dependencies` from `4.2.0` to `4.2.1-SNAPSHOT` in order to pickup spring-cloud/spring-cloud-function#1224 which brings the fix for functions declared as `FactoryBean`.

See #599

### DIFF
--- a/stream-applications-build/pom.xml
+++ b/stream-applications-build/pom.xml
@@ -36,7 +36,9 @@
         <disable.nohttp.checks>true</disable.nohttp.checks>
         <spring-javaformat-checkstyle.version>0.0.43</spring-javaformat-checkstyle.version>
         <spring-boot.version>3.4.2-SNAPSHOT</spring-boot.version>
-        <spring-cloud.version>2024.0.1-SNAPSHOT</spring-cloud.version>
+        <spring-cloud.version>2024.0.0</spring-cloud.version>
+        <!-- SC bom override to pickup spring-cloud/spring-cloud-function#1224 -->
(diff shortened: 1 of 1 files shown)


artembilan:
Not related to the subject, but still good to see what is going on with our new tool DCO.

I think this is not very legit, Chris:
```
Signed-off-by: Chris Bono <cbono@vmware.com>
```
Such an email does not exist any more.
Plus we have discussed that org members no need this `-s`.
However, it turned out that we need GPG signature for our commits instead. 
In the end it might be just personal preferences: `-s` could be very handy when you contribute to external projects 😅.
No need to do anything: just thinking aloud to share my pain making SSH signature today for my commits .

onobc:
> Not related to the subject, but still good to see what is going on with our new tool DCO.
> 
> I think this is not very legit, Chris:
> 
> ```
> Signed-off-by: Chris Bono <cbono@vmware.com>
> ```
> 
> Such an email does not exist any more. Plus we have discussed that org members no need this `-s`. However, it turned out that we need GPG signature for our commits instead. In the end it might be just personal preferences: `-s` could be very handy when you contribute to external projects 😅. No need to do anything: just thinking aloud to share my pain making SSH signature today for my commits .

Ahh yeh, good catch. This was already set in intellij "Sign off commit" so was adding it automatically. i did the same thing in https://github.com/spring-cloud/stream-applications/pull/608

Let me remove these from commit message (in both) and add GPG signature before merge. I will do that tomorrow though.

PR #612, opened 2025-01-23
Update to spring-cloud-function 4.2.1

onobc:
This updates spring-cloud-function to 4.2.1 in order to pickup spring-cloud/spring-cloud-function#1224
which brings the fix for functions declared as `FactoryBean`.


stream-applications-build/pom.xml
```diff
@@ -38,12 +38,12 @@
         <spring-boot.version>3.4.2-SNAPSHOT</spring-boot.version>
```
onobc, line 38: I will update this to just released `3.4.2` version of Boot in a subsequent PR.

PR #619, opened 2025-01-23
Update ivy version to 2.5.2

onobc:
Updates `org.apache.ivy:ivy` to version 2.5.2 to fix CVE-2022-46751


applications/processor/groovy-processor/pom.xml
```diff
@@ -14,10 +14,6 @@
         <relativePath>../../stream-applications-core/pom.xml</relativePath>
     </parent>
 
-    <properties>
-        <apache-ivy.version>2.5.1</apache-ivy.version>
```
artembilan, line 18: Sorry, Chris, how does it work if we don’t specify version any more?
Is this Spring Boot managed dep?
So, it comes as latest now and with all those CVEs fixed?

onobc: Good point @artembilan - I was not clear in my description. We (stream-applications) already have dep. mgmt. for this ivy in [stream-applications-build](https://github.com/spring-cloud/stream-applications/blob/e9afea157b1a8e3e83a2de1fa250d70111594b05/stream-applications-build/pom.xml#L108-L112) but for some reason these versions were still being hardcoded in these modules. 
This moves those versions out of the way and lets the stream-applications-build/pom.xml be the final arbiter of the versions.

.trivyignore
```diff
@@ -1,2 +1,17 @@
+################################
+## From debezium-supplier
+################################
+CVE-2023-1428
```
onobc, line 4: These "mute" the `grpc` and `protobuf` related CVEs being reported

scan-jar.sh
```diff
@@ -4,7 +4,6 @@ SCDIR=$(realpath $SCDIR)
 if [[ "$1" != *"-sources.jar" ]] && [[ "$1" != *"-javadoc.jar" ]]; then
     if [ "$TRIVY_UPLOAD" == "true" ]; then
       echo "Scanning:$1"
-      echo "trivy rootfs --format sarif -o \"$1.sarif\" \"$1\""
```
onobc, line 7: This looks like some debug that is surely to get out of sync w/ the actual command

onobc:
@artembilan I am going to go ahead and merge this.  I am happy to make any changes post merge.

PR #620, opened 2025-01-24
Update Groovy to 3.0.23

onobc:
This updates `org.codehaus.groovy:groovy-all` used by the `stream-applications-release-train` module to `3.0.23` to fix `CVE-2022-4065` from transitive depepdency
`org.testng:testng`.

### DIFF
--- a/stream-applications-release-train/stream-applications-descriptor/pom.xml
+++ b/stream-applications-release-train/stream-applications-descriptor/pom.xml
@@ -15,12 +15,13 @@
     <packaging>jar</packaging>
     <properties>
         <uniqueVersion>false</uniqueVersion>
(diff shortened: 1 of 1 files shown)


stream-applications-release-train/stream-applications-descriptor/pom.xml
```diff
@@ -15,12 +15,13 @@
     <packaging>jar</packaging>
     <properties>
         <uniqueVersion>false</uniqueVersion>
+        <groovy.version>3.0.23</groovy.version>
```
artembilan, line 18: This is very out-dated artifact: https://groovy.apache.org/download.html.
Good to see it is working with whatever Spring Integration does with Groovy!
But probably better to upgrade this artifact group to a newer version.

PR #625, opened 2025-02-01
Fix couple of CVEs before releasing 2024.0.1

onobc:


artembilan:
Why did mark it as duplicate ?

onobc:
> Why did mark it as duplicate ?

Are you talking about the first commit that is the same one that is already in HEAD? I am not sure why Github keeps seeing that as I have rebased etc..

artembilan:
Just label. Removed it. Merging

onobc:
I see it now. Yeh, I had accidentally added that label. Thanks @artembilan

PR #632, opened 2025-05-22
Update to prometheus-rsocket-proxy 2.0.0-RC1

onobc:


onobc:
cc: @jonatan-ivanov

applications/stream-applications-core/pom.xml
```diff
@@ -18,7 +18,7 @@
     <properties>
         <stream-apps-core.version>5.1.1-SNAPSHOT</stream-apps-core.version>
         <java-functions.version>5.1.0</java-functions.version>
-        <prometheus-rsocket.version>1.5.3</prometheus-rsocket.version>
+        <prometheus-rsocket.version>2.0.0-RC1</prometheus-rsocket.version>
```
artembilan, line 21: I wonder if it is really save to do such an upgrade in patch release...

artembilan (changes requested):

